PR #3, opened 2015-09-10
Add MsgDialog headers

SMOKE5:


include/psp2/message_dialog.h
```diff
@@ -0,0 +1,200 @@
...
+enum {
+	SCE_MSG_DIALOG_SYSMSG_TYPE_INVALID					= 0,
+
+	//! Displays "Please wait."
+	SCE_MSG_DIALOG_SYSMSG_TYPE_WAIT						= 1,
+
+	//! Displays "There is not enough free space on the memory card."
+	SCE_MSG_DIALOG_SYSMSG_TYPE_NOSPACE					= 2,
+
+	//! Displays "Move away from the source of interface, or adjust the compass by moving your PS Vita system as shown below."
```
xyzz: interference?

xyzz:
both files are missing license headers

include/psp2/message_dialog.h
SMOKE5: It's a dialog message used with some apps to re calibrate the motion sensors. No idea how that could be used in a homebrew, but I included it anyways

xyzz: I mean your comment says "interface" instead of what I assume to be "interference", just a minor typo.

173210:
Shouldn't you use "common_dialog.h" and "msg_dialog.h" for consistency?
Or you can make a new directory named "dialog" and put "common.h" and "msg.h" in the directory.

xerpi:
Yeah I agree with @173210. Both ways look fine to me.

SMOKE5:
I don't see why it matters since message_dialog.h includes common_api.h, but I'll change it I guess.

173210:
In PSP2SDK, I avoid to include other header files in a header file because it can make invisible dependencies in actual source files which include the header file.
For instance, I caught errors when I reordered `#include` in source files of VHL, because those header files had such dependencies.

You can simply delete those in `#include`. However, in this case, the SceCommonDialog APIs may often be used with SceMsgDialog APIs, so you may have another option to create "dialog.h", which have both APIs.

SMOKE5:
They are seperate libs, but need each other. ( see libSceCommonDialog_stub & libSceAppUtil_stub )
The message dialog stuff is in libSceAppUtil.
So either its done simply this way and includes common_dialog.h or the user will need to do it.
Either way results in the same thing, so I don't see a problem.

xyzz:
Sorry I didn't merge it earlier because I didn't know you fixed the issues I noted, since github only sends emails when comments are added please next time add a comment saying you fixed the issues. Thanks for your pull request.

173210:
@SMOKE5 Some may include `message_dialog.h` to use SceCommonDialog APIs. SceCommonDialog APIs doesn't seem to depend on SceAppUtil, so it's not good.

SMOKE5:
It works either way so no sense in fussing about it. Change it then, I don't care.

PR #9, opened 2016-02-06
Add headers for audio decoding in several formats

SMOKE5:
I have not been able to test SceAtrac, but these should be fine.
The problem is UVLoader cannot fully resolve SceAtrac, all functions return 0xFFFFFFFF. UVLoader doesn't complain when running sceSysmoduleLoadModule(SCE_SYSMODULE_ATRAC); but still, we cannot use the functions currently.


xyzz:
Can you report the issue with uvloader not being able to resolve these functions to https://github.com/yifanlu/UVLoader/issues? Thank you.

TheOfficialFloW:
Hi guys, I've just fixed UVLoader and it can now correctly resolve these NIDs

Issue #15, opened 2016-08-05
libSceSsl_stub missing .h file


InvalidExcepti0n:
working with HTTP had some issues, so started looking into it...
I found out SSL need to be initialised  (and I found lib but no .h file)
when I include .h with this 2 functions it worked
Reference: http://www.smartpsp.net/doxygen-doc-en/html/pspssl_8h.html

int sceSslInit(int poolSize);
int sceSslTerm(void);

as I opened lib with hex there is lot more functions but I needed only this ones
so if someone can look into it and make proper .h file

xyzz:
Fixed by #21

PR #21, opened 2016-08-15
Add libSSL definitions

yne:
As requested by https://github.com/vitasdk/vita-headers/issues/15

Note: the `sceSslGetNotAfter` and `sceSslGetNotBefore` should receive an `SceRtcTick*` instead of `uint64_t*` but the current `psp2/rtc.h` does not have the `SceRtcTick*` type, wish should be :
```
typedef struct SceRtcTick {
	SceUInt64 tick;
} SceRtcTick;
```

Should I also create this type, update the rtc.h and #include it ?

### DIFF
--- a/include/psp2/libssl.h
+++ b/include/psp2/libssl.h
@@ -1,0 +1,54 @@
+/**
+ * \file
+ * \brief Header file which defines SSL variables and functions
+ *
+ * Copyright (C) 2015 PSP2SDK Project
+ *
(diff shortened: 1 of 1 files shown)


xyzz:
> Should I also create this type, update the rtc.h and #include it ?

Sure

yne:
Then I'll create another PR to fix the missing RTC typesdef issue.
In the meantime this one is still good.

PR #39, opened 2016-09-05
Changed pixelformat global name to reflect its behaviour.

Rinnegatamante:
Trying to write a screenshot tool i found that framebuffer is not always set at ABGR encoding, same screenshot code for example produced these three screenshots:
http://rinnegatamante.it/site/PCSB00830_94_16721.bmp
https://pbs.twimg.com/media/CrbxanyXEAEY_Km.jpg:large
https://pbs.twimg.com/media/Cri1FVMXYCAeV7P.jpg:large

### DIFF
--- a/include/psp2/display.h
+++ b/include/psp2/display.h
@@ -31,7 +31,7 @@ enum {
 	SCE_DISPLAY_ERROR_NO_PIXEL_DATA		= 0x80290008
 };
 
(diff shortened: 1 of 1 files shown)


xyzz:
as I mentioned on IRC, this will break every single homebrew in existence, as such, not merging it.

PR #41, opened 2016-09-05
Added some macro names for mode field.

Rinnegatamante:


xyzz:
why are these macros? most constants in vitasdk headers are enums

Rinnegatamante:
Changed with enums.

PR #43, opened 2016-09-05
Removed the additions of void* hostMem and uint hostMemSize, they broke things, sorry

FirebirdTA01:
Added void pointer to host memory and unsigned int hostMemSize to SceGxmRenderTargetParams, needed by SceGxmGetRenderTargetMemSizes() and SceGxmRenderTargetGetHostMem()

### DIFF
--- a/include/psp2/gxm.h
+++ b/include/psp2/gxm.h
@@ -1188,6 +1188,8 @@ typedef struct SceGxmRenderTargetParams {
 	uint16_t scenesPerFrame;        //!< The expected number of scenes per frame, in the range [1,#SCE_GXM_MAX_SCENES_PER_RENDERTARGET].
 	uint16_t multisampleMode;       //!< A value from the #SceGxmMultisampleMode enum.
 	uint32_t multisampleLocations;  //!< If enabled in the flags, the multisample locations to use.
+	void *hostMem;                  //!< The host memory to use for the render target CPU data structures.
+	unsigned int hostMemSize;       //!< The size of the host memory block.
 	SceUID driverMemBlock;          //!< The uncached LPDDR memblock for the render target GPU data structures or SCE_UID_INVALID_UID to specify memory should be allocated in libgxm.
 } SceGxmRenderTargetParams;
 
--- a/include/psp2/gxm.h
+++ b/include/psp2/gxm.h
@@ -1188,8 +1188,6 @@ typedef struct SceGxmRenderTargetParams {
 	uint16_t scenesPerFrame;        //!< The expected number of scenes per frame, in the range [1,#SCE_GXM_MAX_SCENES_PER_RENDERTARGET].
 	uint16_t multisampleMode;       //!< A value from the #SceGxmMultisampleMode enum.
 	uint32_t multisampleLocations;  //!< If enabled in the flags, the multisample locations to use.
-	void *hostMem;                  //!< The host memory to use for the render target CPU data structures.
-	unsigned int hostMemSize;       //!< The size of the host memory block.
 	SceUID driverMemBlock;          //!< The uncached LPDDR memblock for the render target GPU data structures or SCE_UID_INVALID_UID to specify memory should be allocated in libgxm.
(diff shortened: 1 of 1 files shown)


xyzz:
Your indentation is broken, can you fix it?

FirebirdTA01:
Absolutely

joshdekock:
@FirebirdTA01 squash the commits and do a force push

FirebirdTA01:
@joshdekock I'm new to git, I'm used to working on independent locally hosted projects. Could you elaborate for me? Sorry.

joshdekock:
There's two ways to do this, the first would be to do a git soft reset to the previous commit and then recommit the first one, and the second would be to use git rebase. The first is probably easier.

```
git reset --soft a9b243a82d8d5c9af00582639c6273eba030c759
git add include/psp2/gxm.h
git commit -m "Added void *hostMem and u int hostMemSize"
```
And then check it all looks ok with `git log`, before updating this PR: `git push --force`

FirebirdTA01:
The comment indentation seems to be a space ahead of the tabbed indent on both new lines now, instead of a tab behind on the first one and correct on the second.  Visual Studio's editor doesn't show this and I've tried reloading files and normalizing line files. Do things look okay or am I still doing something wrong?

FirebirdTA01:
Okay okay, I see that tabs were not used in the original source. I replaced with normal spaces and things appear as they should now. I feel like an idiot, need more coffee! Haha, thanks for the help guys!

joshdekock:
Have you tested it? Can I see a sample of the usage?

FirebirdTA01:
I was working on shaders, ill upload what ive got once im back by my main computer

FirebirdTA01:
It does work, I was having issues with the programs crashing but it was due to my main loop trying to start the drawing methods before the graphics initialization was finished.
Its usage is as follows:
`int error = 0;
//set up render target parameters
memset(&renderTargetParams, 0, sizeof(renderTargetParams));
renderTargetParams.flags = 0;
renderTargetParams.width = SCREEN_WIDTH;
renderTargetParams.height = SCREEN_HEIGHT;
renderTargetParams.scenesPerFrame = 1;
renderTargetParams.multisampleMode = MSAA_MODE;
renderTargetParams.multisampleLocations = 0;
renderTargetParams.driverMemBlock = -1;

//These will be se correctly in the code immediately following
renderTargetParams.hostMem = NULL;
renderTargetParams.hostMemSize = 0;

//calculate correct memory sizes
unsigned int hostMemSize, driverMemSize;
sceGxmGetRenderTargetMemSizes(&renderTargetParams, &hostMemSize, &driverMemSize);
//allocate the host memory
printf("Allocating the host memory, size: %u\n", hostMemSize);
renderTargetParams.hostMem = malloc(hostMemSize);
renderTargetParams.hostMemSize = hostMemSize;

//finally create the render target
error = sceGxmCreateRenderTarget(&renderTargetParams, &gxmRenderTarget_ptr);
printf("sceGxmCreateRenderTarget() result: 0x%08X\n", error);`

FirebirdTA01:
I've noticed something else, I don't think it could be a result of just the changes from this commit but all of the packages built from my local libraries/toolchain now request extended permissions while installing. Could somebody enlighten me as to why this could be happening?

FirebirdTA01:
@joshdekock @xyzz @yifanlu 
I was wondering if any of you could shed some light on some problems I'm having with the sceGxmCreateRenderTarget function. It's returning error 0x805B0003 saying that one of the initialize parameters is incorrect. I've speculated that the cause of this error may be due to the hostMemSize parameter being an unsigned int type because when checking against a "different" sdk, that parameter is a uint32_t type. I've changed both the parameter var type and the argument types for SceGxmGetRenderTargetMemSizes() trying to use the uint32_t type, and the SceSize type since that was successful when creating a gxmContext, but the error returned remains the same.  The hostMem parameter is described as important, to say the least, in some docs and a renderTarget is needed in order to start or end scenes on the GPU. Do you guys have any ideas?

PR #47, opened 2016-09-10
Added missing sdkVersion to param

TheOfficialFloW:


include/psp2/common_dialog.h
```diff
@@ -120,6 +121,7 @@ void sceCommonDialogConfigParamInit(SceCommonDialogConfigParam *param)
 	memset(param, 0x0, sizeof(SceCommonDialogConfigParam));
 	param->language = ~0U;
 	param->enterButtonAssign = ~0U;
+	param->sdkVersion = 0x03150021;
```
DaveeFTW: What is the meaning of this constant?

DaveeFTW:
This change would imply that existing software using affected structure probably should not work; is this the case?

xyzz:
poke @TheOfficialFloW

TheOfficialFloW:
The constant `0x03150021 `dosen't have any meaning for this param, it is just a value that I have used for in dialogs when I reverse engineered the params in PSM (yes, I didn't have any leaked sdk's back then). (For the message dialog, having this sdk version would allow us to have 1024 chars instead of 512 IIRC).

We can define an other sdk version if suggested (3.60 maybe?).

PR #49, opened 2016-09-10
Added sceKernelGetModel signature and NID

himanshugoel2797:


DaveeFTW:
Can you write a small description for this? Would also be useful to provide an enum for the return values, assuming its similar to PSP.

himanshugoel2797:
I can do the description, don't have all the values to be able to do the enum.
On the PCH-1000 it returns 0x10000, don't know about others.

joel16:
Slim returns 0x10000 as well. Vita TV however returns 0x20000.

endrift:
Functions that should not be passed arguments take `(void)`, not `()`. `()` means you can pass any arguments (only in C).

himanshugoel2797:
I know, I thought that it'd be better to stick to just `()` to be consistent with the existing signatures.

himanshugoel2797:
Just added the comments and enum

xyzz:
Ok we can fix the () stuff later (and on the whole repo at once).

PR #52, opened 2016-09-11
Some stuffs (more info in comment)

Rinnegatamante:
- Updated copyright references in every header file.
- Fixed an header guard name in psp2/kernel/loadcore.h.
- Divided MAX_REPORT in psp2/touch.h for FRONT and BACK ports.
- Properly documented psp2/power.h.


xyzz:
you can't just remove copyrights like that. if you really really want, you can leave the psp2sdk line and add a new vitasdk line below it

Rinnegatamante:
Fixed

include/psp2/touch.h
```diff
@@ -18,9 +18,9 @@ extern "C" {
 
 #include <psp2/types.h>
 
-// SCE_TOUCH_MAX_REPORT = 8
 enum {
-	SCE_TOUCH_MAX_REPORT	= 8	//!< FIXME 6 on front | 4 on back
```
DaveeFTW: We can't remove this without breaking homebrew that uses it.

Rinnegatamante: I know but it's just wrong. Not a big deal to just change the variable name in homebrews that use it. Even this got merged and broke tons of homebrews cause sceRtcTime became sceDateTime: https://github.com/vitasdk/vita-headers/commit/2b10387cc0db941a216097dcb631f6e85607b95f

Rinnegatamante:
What about making SCE_TOUCH_MAX_REPORT to be SCE_TOUCH_MAX_REPORT_BACK + SCE_TOUCH_MAX_REPORT_FRONT so that it will not break homebrews using it?

xyzz:
I don't have any idea what your touch.h changes are trying to achieve. This constant is only used in struct from the same touch.h header. Now that you removed it it just won't compile at all. I don't see much point in providing SCE_TOUCH_MAX_REPORT_BACK/SCE_TOUCH_MAX_REPORT_FRONT. Maybe the only use is that a developer can look at it and not make a homebrew that would require more than 6 simultaneous presses?

Anyway, right now you aren't supposed to use SCE_TOUCH_MAX_REPORT at all. What you do is define `SceTouchData`, get touch reports, iterate up to `reportNum`. SCE_TOUCH_MAX_REPORT never comes into play here. I suggest reverting this change.

Rinnegatamante:
What's the point with this so? "//!< FIXME 6 on front | 4 on back"

xyzz:
I don't know.

Rinnegatamante:
Is it fine now or should i remove also the enums values for FRONT and BACK and restore the old situation with the FIXME alert?

joshdekock:
@Rinnegatamante `documenting ScePower` looks like the only mergeable thing at this point in time, if you're happy with just that being merged that I'll go ahead and do that.

Rinnegatamante:
As you wish, i can modify the commits if needed.

joshdekock:
If you could drop the other commits using `git rebase`, and force push that'd be great, otherwise I can do it

PR #56, opened 2016-09-17
Add more SceCtrl NIDs

d3m3vilurr:
* Add more button values; L1/R1/L3/R3
* Add more function defines; 
  - sceCtrlReadBufferPositiveExt2
  - sceCtrlSetSamplingModeExt
  - sceCtrlGetControllerPortInfo
  - sceCtrlGetBatteryInfo
* Add SceCtrlPortInfo struct
* Update SceCtrlData struct


include/psp2/ctrl.h
```diff
@@ -28,13 +28,17 @@ enum {
...
+	SCE_CTRL_L3         = 0x000002,	//!< L3 button.
+	SCE_CTRL_R3         = 0x000004,	//!< R3 button.
 	SCE_CTRL_START      = 0x000008,	//!< Start button.
 	SCE_CTRL_UP         = 0x000010,	//!< Up D-Pad button.
 	SCE_CTRL_RIGHT      = 0x000020,	//!< Right D-Pad button.
 	SCE_CTRL_DOWN       = 0x000040,	//!< Down D-Pad button.
 	SCE_CTRL_LEFT       = 0x000080,	//!< Left D-Pad button.
 	SCE_CTRL_LTRIGGER   = 0x000100,	//!< Left trigger.
 	SCE_CTRL_RTRIGGER   = 0x000200,	//!< Right trigger.
+	SCE_CTRL_L1         = 0x000400,	//!< L1 button.
```
xyzz: L1 or L2?

d3m3vilurr: in sceCtrlReadBufferPositiveExt2, L2 button set 0x100(LTRIGGER) but L1 button set 0x400.
(also R2 set 0x200, R1 set 0x800)
but general used sceCtrlPeekBufferPositive, only can use LT/RT values.

can you suggest better name if it looks confused?

xyzz: so when used with normal functions, `SCE_CTRL_LTRIGGER` is L1 on dualshock and L1 on ps vita; however when used with `sceCtrlReadBufferPositiveExt2` `SCE_CTRL_LTRIGGER` is L2 on dualshock and does not appear at all on ps vita?

I guess if I got this right, you should specify for _L1/_R1 that these are only used for the `sceCtrlReadBufferPositiveExt2` function

d3m3vilurr: > so when used with normal functions, SCE_CTRL_LTRIGGER is L1 on dualshock and L1 on ps vita; 

yes

> however when used with sceCtrlReadBufferPositiveExt2 SCE_CTRL_LTRIGGER is L2 on dualshock and does not appear at all on ps vita?

actually psvita not have l2/r2/l3/r3 buttons, so not appear :) it look like swap.

> I guess if I got this right, you should specify for _L1/_R1 that these are only used for the sceCtrlReadBufferPositiveExt2 function

well...... `LTRIGGER` is can confuse.
in many cases(like xbox, ds, etc), `LTRIGGER` is not mean `L1`, mean L2 button.
but we use expect `L1`

How about define all values ? like;
```
{
  ...
  SCE_CTRL_LTRIGGER = 0x000100,
  SCE_CTRL_L1 = 0x000400, // or SCE_CTRL_EXT_L1 = 0x000400,
  SCE_CTRL_L2 = 0x000100, // duplicate define but...
  SCE_CTRL_L3 = 0x000002,
...
}
```

or want you another enum set? like;

```
{
  SCE_CTRL_EXT_L1 = 0x000400,
  SCE_CTRL_EXT_L2 = 0x000100,
  SCE_CTRL_EXT_L3 = 0x000002,
}
```
but this is little strange, maybe we use both enums at same time, we need to split?

and another options, define L2 to 0x000400, then make compatible user method, like `sceCtrlPeekBufferPositiveExtend`..

include/psp2/ctrl.h
```diff
@@ -191,6 +218,13 @@ int sceCtrlSetActuator(int port, const SceCtrlActuator* pState);
  */
 int sceCtrlSetLightBar(int port, SceUInt8 r, SceUInt8 g, SceUInt8 b);
 
+/**
+ * Get controller port information.
+ *
+ * @param[out] info - Return value, use char[16] buffer
+ * @return 0, <0 on error
+ */
+int sceCtrlGetControllerPortInfo(void *info);
```
DaveeFTW: Any more information for this function? I understand from the name that it gets the controller port information but can you clarify what this information is? stuff like format, etc

d3m3vilurr: I don't know detail PortInfo struct (so i add just void*)

```
uint8_t info[16];
int ret = sceCtrlGetControllerPortInfo(info);
if (ret < 0) {
  // error
}
if (port < 0 || port > 4) { // port range 0~4
  // error
}
switch(info[port]) {
  case 1:
    // maybe psvita
  case 2:
    // i was only seen index 0 this value
  case 4:
    // maybe ds4 pad
  case 8:
    // 
  case 0:
    // maybe disconnected port
}
```
other bytes currently unknown...
maybe first byte mean connected port count or bit flags, but don't know yet
(i need detail test result with multiple pad, but i haven't pstv)

value return examples;
```
PSVITA:           0x01 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00
PSTV with DS4 1): 0x02 0x04 0x00 0x00 0x00 0x0D 0x18 0x81 0x10 0x69 0x00 0x00 0x38 0x00 0x0C 0x81
PSTV with DS4 1): 0x02 0x04 0x00 0x00 0x00 0x0C 0x10 0x81 0x10 0x69 0x00 0x00 0x38 0x00 0x04 0x81
```

d3m3vilurr:
@DaveeFTW @xyzz I push 1 more commit to this PR.
I found sceCtrlGetBatteryInfo & LT/RT analog fields.
Please check last commit.

PR #57, opened 2016-09-18
add sceAppMgrGetAppParam

SMOKE5:
This function copies the app parameters to an array through which you can grab custom arguments from, kind of a substitute for argv

It returns something like this if you use it from a custom uri:

type=LAUNCH_APP_BY_URI&uri=test:test?test

### DIFF
--- a/include/psp2/appmgr.h
+++ b/include/psp2/appmgr.h
@@ -83,6 +83,9 @@ int _sceAppMgrGetAppState(SceAppMgrAppState *appState, uint32_t len, uint32_t ve
 
 int sceAppMgrReceiveSystemEvent(SceAppMgrSystemEvent *systemEvent);
 
+//! Copies app param to an array
+int sceAppMgrGetAppParam(char *param);
+
 //! Obtains the BGM port, even when it is not in front
 int sceAppMgrAcquireBgmPort(void);
 
--- a/include/psp2/appmgr.h
+++ b/include/psp2/appmgr.h
@@ -84,6 +84,7 @@ int _sceAppMgrGetAppState(SceAppMgrAppState *appState, uint32_t len, uint32_t ve
 int sceAppMgrReceiveSystemEvent(SceAppMgrSystemEvent *systemEvent);
 
 //! Copies app param to an array
+//! App param example: type=LAUNCH_APP_BY_URI&uri=psgm:play?titleid=NPXS10031
 int sceAppMgrGetAppParam(char *param);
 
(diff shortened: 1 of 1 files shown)


DaveeFTW:
I think adding your example URI call to the header would be very beneficial.

SMOKE5:
I added a psgm uri (for PKG Installer) since people will be more familiar with that one

DaveeFTW:
Great, lastly, can you give add comments on the return type?

SMOKE5:
added, though it should never return anything but 0

SMOKE5:
I don't think this is worth making a sample of, but here's some code to show how it works

`char AppParam[1024];

sceAppMgrGetAppParam(AppParam);
printf("arg = %s", AppParam);`

endrift:
Do you not pass the size of the array as a second parameter? Does it assume a fixed size?

SMOKE5:
I don't think so, @endrift but maybe you can double check? I'm not the best at RE. Look in the browser

TheOfficialFloW:
The function is correct. There's no second argument.

include/psp2/appmgr.h
```diff
@@ -83,6 +83,11 @@ int _sceAppMgrGetAppState(SceAppMgrAppState *appState, uint32_t len, uint32_t ve
 
 int sceAppMgrReceiveSystemEvent(SceAppMgrSystemEvent *systemEvent);
 
+//! Copies app param to an array
+//! App param example: type=LAUNCH_APP_BY_URI&uri=psgm:play?titleid=NPXS10031
+//! Returns 0 for success
+int sceAppMgrGetAppParam(char *param);
```
DaveeFTW: should this be a ``const char *`` instead?

endrift: N...no? That's an output buffer. This is *Get*AppParam.

endrift:
Do we know the size the output buffer needs to be?

include/psp2/appmgr.h
DaveeFTW: Yeah, I'm being stupid.

TheOfficialFloW:
@endrift, 1024bytes according to SceSettings.

endrift:
That should probably be documented in the comment...on the PR that just got merged, oops.

DaveeFTW:
@endrift @TheOfficialFloW I've updated this in a follow-up PR https://github.com/vitasdk/vita-headers/pull/59

endrift:
Thanks.

PR #46, opened 2016-09-10
Add function signature for _sceAppMgrDestroyAppByName

himanshugoel2797:

PR #58, opened 2016-09-21
Fixed number of arguments in two functions

TheOfficialFloW:
The number of arguments in the functions commited by @himanshugoel2797 were wrong.

@himanshugoel2797, reverse enginering the api properly instead of just guessing them.

### DIFF
--- a/include/psp2/appmgr.h
+++ b/include/psp2/appmgr.h
@@ -75,7 +75,8 @@ typedef struct SceAppMgrLaunchAppOptParam SceAppMgrLaunchAppOptParam; // Missing
 int sceAppMgrDestroyAppByName(char *name);
 
 //! name: The Title ID of the application
-int _sceAppMgrDestroyAppByName(char *name, char temp[3]);
+//! temp: A 12bytes temporary work buffer (content important)
+int _sceAppMgrDestroyAppByName(const char *name, char temp[3]);
 
 int _sceAppMgrGetAppState(SceAppMgrAppState *appState, uint32_t len, uint32_t version);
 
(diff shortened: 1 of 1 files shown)


include/psp2/appmgr.h
```diff
@@ -67,11 +67,15 @@ typedef struct SceAppMgrSystemEvent {
...
+typedef struct SceAppMgrLaunchAppOptParam SceAppMgrLaunchAppOptParam; // Missing struct
 
 #define SCE_APPMGR_MAX_APP_NAME_LENGTH	(31)
 
 //! name: The Title ID of the application
-int _sceAppMgrDestroyAppByName(char *name);
+int sceAppMgrDestroyAppByName(char *name);
+
+//! name: The Title ID of the application
+int _sceAppMgrDestroyAppByName(char *name, char temp[3]);
```
DaveeFTW: I don't know any API using temporary values in parameters. Can you confirm this or add an appropriate comment for other developers? Also, can you confirm the const-ness of the name parameter?

include/psp2/appmgr.h
```diff
@@ -103,7 +107,7 @@ int sceAppMgrLoadExec(const char *appPath, char * const argv[],
 int sceAppMgrLaunchAppByUri(int flags, char *uri);
 
 //! name: The Title ID of the application
-int sceAppMgrLaunchAppByName2(char *name);
+int sceAppMgrLaunchAppByName2(char *name, char *param, SceAppMgrLaunchAppOptParam *optParam);
```
DaveeFTW: Add comments for these new parameters and confirm const-ness.

include/psp2/appmgr.h
TheOfficialFloW: Nearly every _sceAppMgr has got a temporary work buffer which are passed in the wrapper module SceDriverUser. Using sceAppMgrDestroyAppByName wouldn't require to set this temporary buffer. Idecided to still keep it, because the latest VitaShell uses this API.

DaveeFTW: Is there any benefit to using the ``_sceAppMgrXX`` API over its non-underscore counterpart?

TheOfficialFloW: There aren't. It's the only `_sceAppMgrXX` function here, because of the reason I mentioned about VitaShell. Or should I rather update VitaShell's code and remove it?

DaveeFTW: I think for consistency we should keep the SDK using the non-underscore API. It gives us two benefits:
- Consistency within the SDK
- Theoretical API protection. I assume that the underscore API is private/hidden from developers and thus can be changed. This of course, may be completely false.

This would be effectively reverting https://github.com/vitasdk/vita-headers/pull/46 which can break homebrew other than VitaShell. @xyzz @joshdekock any input?

joshdekock: I think if we can avoid adding any more underscore functions that would be preferred. We should add inline wrappers without underscores for functions which do not have native underscore-less wrappers.

joshdekock:
Pinging @TheOfficialFloW on an update to this.

PR #62, opened 2016-09-30
Added location.h for gps/3g/wifi location services

FirebirdTA01:


FirebirdTA01:
I wrote the file just a day ago, I miswrote the copywrite info when I pushed the file

PR #63, opened 2016-10-01
Update SceCtrlActuator

d3m3vilurr:
`enable` field split to `small` and `large`
These fields mean motor strength, and can be 0 to 0xff

### DIFF
--- a/include/psp2/ctrl.h
+++ b/include/psp2/ctrl.h
@@ -102,8 +102,9 @@ typedef struct SceCtrlRapidFireRule {
 
 /** Structure to pass as argument to ::sceCtrlSetActuator */
 typedef struct SceCtrlActuator {
-	int enable;  //!< Enable the actuator vibration
-	int unk;     //!< Unknown
+	unsigned char small; //!< Vibration strength of the small motor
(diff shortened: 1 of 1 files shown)


include/psp2/ctrl.h
```diff
@@ -102,8 +102,9 @@ typedef struct SceCtrlRapidFireRule {
 
 /** Structure to pass as argument to ::sceCtrlSetActuator */
 typedef struct SceCtrlActuator {
-	int enable;  //!< Enable the actuator vibration
-	int unk;     //!< Unknown
+	unsigned char small; //!< Strength of the small motor vibration
+	unsigned char large; //!< Strength of the large motor vibration
+	uint8_t unk[8]; //!< Unknown
```
xyzz: why is the struct different size now?

d3m3vilurr:
More detail info;

DS4:
all fields work 0x0 to 0x255
DS3:
`small` field always work max strength
`large` field work 0x40~0xff

I guess, DS3 maybe use lower spec vibration moters.

include/psp2/ctrl.h
d3m3vilurr: huh mistype i will fix them.

PR #65, opened 2016-10-11
Use enum tag in return types gxt.h.

HybridEidolon:
When using rust bindgen to generate Rust bindings over the psp2 headers, clang encounters a compile error in using an enum as a type without the `enum` tag. Evidently this is a semantic error; you are _supposed_ to use `enum` tags, but gcc allows not using it without warning.


d3m3vilurr:
why not add just typedef?

PR #67, opened 2016-11-03
kernel development headers

DaveeFTW:


yifanlu:
The main change here is that I moved the kernel libraries out of SceLibKernel and into their own libraries. The new toolchain will add those kernel modules stubs by default.

PR #68, opened 2016-11-03
Added some SceShellUtil functions

TheOfficialFloW:
These have been reverse engineered and are used lock/unlock the PS button or others

### DIFF
--- a/include/psp2/shellutil.h
+++ b/include/psp2/shellutil.h
@@ -1,0 +1,83 @@
+/**
+ * \file
+ * \brief Header file which defines shell utils
+ *
+ * Copyright (C) 2016 Vitasdk developers
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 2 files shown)


Banandana:
Thank you for this :+1:

include/psp2/shellutil.h
```diff
@@ -0,0 +1,83 @@
+/**
+ * \file
+ * \brief Header file which defines shell utils
+ *
+ * Copyright (C) 2015 PSP2SDK Project
```
TheOfficialFloW: ok..changed

DaveeFTW:
Unfortunately, you've been made an example of here. We're probably going to see more contributions with functions that don't have any official names and this is a good opportunity to enforce some sort of "standard" for contribution. Sorry.

Names such as ```SceShellUtil_E8AD11EC``` and ```SceShellUtil_7B5EDFE7``` are not helpful to developers. If there are no official names available then use intuition and create suitable names.

If you check the headers in ```psp2kern/``` most of those are unofficial names but attempt to follow the styling of the official ones.

Otherwise, I am happy to merge this again.

PR #76, opened 2016-11-05
Fix PR #74

d3m3vilurr:
Rebase from current master tree.
Conflict introduced #71 but @TheOfficialFloW not merge that.

### DIFF
--- a/db.json
+++ b/db.json
@@ -4708,6 +4708,22 @@
         }, 
         "nid": 202
     }, 
+    "SceShellSvc": {
+        "modules": {
+            "SceShellUtil": {
+                "functions": {
+                    "sceShellUtilRegisterEventHandler": 2069815271, 
+                    "sceShellUtilInitEvents": 3903656428, 
+                    "sceShellUtilLock": 2840818227, 
+                    "sceShellUtilUnlock": 564580180
+                }, 
+                "kernel": false, 
+                "nid": 3534866606, 
+                "variables": {}
+            }
+        }, 
+        "nid": 2182988835
+    }, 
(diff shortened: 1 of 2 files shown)


d3m3vilurr:
```
23:54:10 F<frangarcj> can i merge pr 76?
23:56:22 X<xyz[m]> frangarcj: sceShellUtilInitEvents - earlier version noted that you have to call this before calling lock, is that still the case?
23:59:18 D<d3m3vilurr> xyz[m]: i guess still need it.
23:59:20 D<d3m3vilurr> https://github.com/TheOfficialFloW/VitaShell/commit/b7005c0bd5103440359c236877dfab1f76947e37
23:59:28 D<d3m3vilurr> https://github.com/TheOfficialFloW/VitaShell/commit/0e9e87fa16b8e43ecf1fbc1abbec0f78c62f4865
23:59:54 D<d3m3vilurr> flow's vitashell patch still have sceShellUtilInitEvents
00:01:12 F<frangarcj> what's the problem?
00:01:51 D<d3m3vilurr> i guess more generic name is `sceShellUtilInit` or `sceShellUtilServiceInit` instead that name.
```
@TheOfficialFloW what your opinion about this discussion?

TheOfficialFloW:
This function has been seen in ScePspemu only, so I wouldn't use a generic name for it. **sceShellUtilInit** or **sceShellUtilServiceInit** would imply that it is required for any sceShellUtil functions which isn't the case.
Everything we know about this function is that it is required for events lock/unlock, so let's call it 'sceShellUtilEventInit' then.

PR #79, opened 2016-11-07
Add NpDrmPackage

devnoname120:
Reversed and documented by @173210.
See http://173210.github.io/scenpdrmpackage

I'm not sure about the rule for prefixing function names. Should I remove the `_`?

### DIFF
--- a/db.json
+++ b/db.json
@@ -3552,6 +3552,20 @@
         }, 
         "nid": 119
     }, 
+    "SceNpDrmPackage": {
+        "modules": {
+            "SceNpDrmPackage": {
+                "functions": {
+                    "_sceNpDrmPackageCheck": 2715321850,
+                    "_sceNpDrmPackageDecrypt": 3606076108
+                },
+                "kernel": false,
+                "nid": 2287029682,
+                "variables": {}
+            }
+        },
+        "nid": 120
+    },
     "SceNpManager": {
         "modules": {
             "SceNpManager": {
--- a/include/psp2/npdrmpackage.h
+++ b/include/psp2/npdrmpackage.h
@@ -2,7 +2,7 @@
  * \file
  * \brief Header file which defines NpDrmPackage related variables and functions
  *
- * Copyright (C) 2015 PSP2SDK Project
+ * Copyright (C) 2016 PSP2SDK Project
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
@@ -39,6 +39,8 @@ typedef struct {
  * @param identifier - arbitrary value [0, 6) ORed with (1 << 8) or 0.
  *                     If it is set to 0, the function just checks the header
  *                     and doesn't create the context.
+ *
+ * @return 0 on success, != 0 on error
  */
 int _sceNpDrmPackageCheck(const void *buffer, SceSize size, int zero, unsigned int identifier);
 
@@ -48,6 +50,8 @@ int _sceNpDrmPackageCheck(const void *buffer, SceSize size, int zero, unsigned i
  * @param buffer - The buffer containing the content of the PKG.
  * @param size - The size of the buffer. The minimum confirmed value is 0x20.
  * @param opt - The options.
+ *
+ * @return 0 on success, != 0 on error
  */
 int _sceNpDrmPackageDecrypt(void * restrict buffer, SceSize size, _sceNpDrmPackageDecrypt_opt * restrict opt);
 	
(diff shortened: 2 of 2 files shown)


yifanlu:
Yeah, remove the "_" if there's colliding names, the convention is to add "ForKernel" or "ForDriver" to ones in kernel.

173210:
In my opinion:
> I think it is legitimate because they are intended for the internal use. Those functions are actually primitive and a userland static library gives more advanced functions. Probably they are expected to be called just by the library and users will use the library instead of calling them by themselves.

Note that the library is on userland but not on kernel. Anyway, I don't think the prefix really matters in the actual use.

d3m3vilurr:
generically, `_` mean `private` or `internal function`.
so i also agree yifanlu`s opinion.
how about just remove `_`?

devnoname120:
Many functions from the pastebin file seem to lose one `_` when added to `db.json`. If they are prefixed with `__`, then it becomes `_`, and `_` becomes nothing. But this is not the case for all of them.

@yifanlu Some rules should be defined. There are currently many discepancies in `db.json`.

173210:
I think that is because those names are cited from the binary of Sony. However, some are determined in the community and they cannot be distinguished from the "real" names. It is no longer possible and doesn't make sense to track real names. We should have our own rules.

devnoname120:
@yifanlu Let's merge it as is, it will be fixed with the rest once the standard is defined.

devnoname120:
This pull request is broken since the changes have been overwritten by others. I'll try to recover the changes.

devnoname120:
Done.

include/psp2/npdrmpackage.h
```diff
@@ -0,0 +1,58 @@
...
+ * Read the header of the PKG and initialize the context
+ *
+ * @param buffer - The buffer containing the header of PKG.
+ * @param size - The size of buffer. The minimum confirmed value is 0x8000.
+ * @param zero - Unknown. Supposed to be set to 0.
+ * @param identifier - arbitrary value [0, 6) ORed with (1 << 8) or 0.
+ *                     If it is set to 0, the function just checks the header
+ *                     and doesn't create the context.
+ */
+int _sceNpDrmPackageCheck(const void *buffer, SceSize size, int zero, unsigned int identifier);
```
DaveeFTW: Can you add documentation for the return value?

include/psp2/npdrmpackage.h
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Decrypt a PKG
+ *
+ * @param buffer - The buffer containing the content of the PKG.
+ * @param size - The size of the buffer. The minimum confirmed value is 0x20.
+ * @param opt - The options.
+ */
+int _sceNpDrmPackageDecrypt(void * restrict buffer, SceSize size, _sceNpDrmPackageDecrypt_opt * restrict opt);
```
DaveeFTW: same here

include/psp2/npdrmpackage.h
```diff
@@ -0,0 +1,58 @@
+/**
+ * \file
+ * \brief Header file which defines NpDrmPackage related variables and functions
+ *
+ * Copyright (C) 2015 PSP2SDK Project
```
DaveeFTW: do you mean 2015?

db.json
```diff
@@ -3552,6 +3552,20 @@
...
+                "functions": {
+                    "_sceNpDrmPackageCheck": 2715321850,
+                    "_sceNpDrmPackageDecrypt": 3606076108
+                },
+                "kernel": false,
+                "nid": 2287029682,
+                "variables": {}
+            }
+        },
+        "nid": 120
```
DaveeFTW: need module nid

include/psp2/npdrmpackage.h
devnoname120: Is this date thing really necessary? It seems to be an unnecessary burden, although minor.

include/psp2/npdrmpackage.h
devnoname120: I suppose it's < 0 on error, otherwise 0. Need @173210 's confirmation.

db.json
DaveeFTW: Most do, yes. We have the technology to fill these in properly now.

include/psp2/npdrmpackage.h
173210: We should follow the existing code anyway.

include/psp2/npdrmpackage.h
173210: != 0 on error, otherwise 0.

db.json
devnoname120: I think that this should be done with the massive overhault, including the _ change.

devnoname120:
Done the changes except `_`.

include/psp2/npdrmpackage.h
devnoname120: OK.

PR #82, opened 2016-11-10
Add sqlite3.h, fix variables set as function in db.json

devnoname120:
I ran the following code on 3.60 to get the version:
```C
#include <psp2/sysmodule.h>
#include <debugnet.h>

const char *sqlite3_libversion(void);

int main()
{
	debugNetInit(YOURPCIPHERE, 18194, DEBUG);

	sceSysmoduleLoadModule(SCE_SYSMODULE_SQLITE);
	const char *ver = sqlite3_libversion();

	debugNetPrintf(DEBUG, "%s\n", ver);
}
```
I then downloaded `sqlite3.h` from here: https://www.sqlite.org/sqlite-amalgamation-3.6.23.1.tar.gz

In order to figure out the variables that were incorrectly set as function in `db.json`, I used a simple script that looked for every entry in the header file to determine whether it's a function or a variable.

### DIFF
--- a/db.json
+++ b/db.json
@@ -4940,7 +4940,6 @@
                     "sqlite3_step": 3397866935, 
                     "sqlite3_stmt_status": 4155241978, 
                     "sqlite3_strnicmp": 316865560, 
-                    "sqlite3_temp_directory": 451682164, 
                     "sqlite3_test_control": 843927293, 
                     "sqlite3_thread_cleanup": 389848075, 
                     "sqlite3_threadsafe": 2746751194, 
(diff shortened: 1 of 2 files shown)


devnoname120:
Please let me rebase before merging.

devnoname120:
I can't test it thoroughly as the following code displays `out of memory`.

```C
#include <sqlite/sqlite3.h>
#include <psp2/sysmodule.h>


sceSysmoduleLoadModule(SCE_SYSMODULE_SQLITE);
sqlite3* db;
int ret = sqlite3_open("ur0:shell/db/app.db", &db);
if (ret != 0) {
    printf(sqlite3_errmsg(db));
    sqlite3_close(db);
}
```

I suspect that this issue would be resolved by defining `sqlite3_temp_directory`, but it's set as a function in the vitasdk, and I can't get the vitasdk to compile so that I can try my change to `db.json`.

devnoname120:
(To explain the above, the pull request closed by itself because I did a `git reset` to fix the branch, so it had the same state as master.)

Rinnegatamante:
sqlite3_temp_directory must be always defined by user with SceSqlite since it isn't already defined by default (like standard sqlite3).

devnoname120:
Outdated.

PR #86, opened 2016-11-12
[WIP] Add NIDs from other sources

xerpi:
* wiki.henkaku.xyz
* "Vita NIDs pastebin"
* prettify db

Info: used https://gist.github.com/xerpi/eb49b34f7bc43f90194c0a16bbf0e623 to "link" the jsons DB together


d3m3vilurr:
well... i'm following commits, have any reason about change tab char in 276c2e2? it make to change all lines, then cannot check diff..

d3m3vilurr:
anything else LGTM. need you more personal check (like duplicate value check using tool)? if need, i'll make checking script.

xerpi:
276c2e2 sorts all the entries, so it's very difficult to see the changes easily anyways.

d3m3vilurr:
if one commit have only one action(in this case i just ignore that commit)
but maybe they still can be diff.
but indent change can't, if reason is match other header files (they use hardtab), i can allow that.
but another reason like reduce file size, i can't understand. hm...but yeah, it's little things
so i'm ok this patch, but i'll make checking script for the automatic compare old db values and entry duplication.
or if you already checked thase, please notice to me. i'll just merge this.

xerpi:
Do not merge yet, there's a a nid duplication problem:
<xerpi> why the hell did Sony put some sceIo functions to SceLibKernel?
<xerpi> on 1.69 those functions are on SceIofilemgr
<xerpi> but on 3.60 for example, they're in SceLibKernel
<xerpi> same with threadmgr functions

d3m3vilurr:
https://gist.github.com/d3m3vilurr/034169e6ea137bad24b8cf433cdd93a6 quick&dirty check.
@xerpi yeah.. it's weird :(

xerpi:
The problem is that I added NIDs from the wiki, but those NIDs are the 1.69 ones, and some of them are in another module now on 3.x.

db.json
```diff
@@ -2503,7 +2503,6 @@
 					"sceKernelCancelMsgPipe": 3463879811,
 					"sceKernelCancelMutex": 460639113,
 					"sceKernelCancelMutex_089": 1912911942,
```
devnoname120: What does this suffix mean?

yifanlu:
We should also get module nids from the wiki. Since module nids change per fw version, we use 1.69 if possible. Then, optionally, we can use the yaml overlay to add module nids for other fw versions.

devnoname120:
This pull request is outdated since we switched to YAML.  In addition, it has unresolvable conflicts because it has been sorted.

Someone should create a new pull request and post the code (preferably Python) to produce each step, so that we can proof-check the code rather than the result.

devnoname120:
And so that we don't forget about this: #86.

PR #96, opened 2016-11-17
Add travis settings for autogenerate documents

d3m3vilurr:
if pushed into master branch, automatically regenerate gh-pages branch.

this patch require to setup `$GH_TOKEN` env var to travis settings.

### DIFF
--- a/.travis.d/deploy_docs.sh
+++ b/.travis.d/deploy_docs.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+cd $TRAVIS_BUILD_DIR
+git fetch origin gh-pages:origin/gh-pages
+git checkout origin/gh-pages -b gh-pages || git checkout --orphan gh-pages
+git reset
+ls | grep -v 'docs' | xargs rm -rf
+mv docs/html/* .
+rm -rf docs
+git add *
+git commit -am "Generated documents from TravisCI for build $TRAVIS_BUILD_NUMBER"
+git push -q https://$GH_TOKEN@github.com/$TRAVIS_REPO_SLUG.git gh-pages:gh-pages -f 2> /dev/null
--- a/.travis.d/deploy_docs.sh
+++ b/.travis.d/deploy_docs.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 set -e
 
-cd $TRAVIS_BUILD_DIR/docs && doxygen
+cd $TRAVIS_BUILD_DIR/docs
+$HOME/doxygen/doxygen*/bin/doxygen
 
 cd $TRAVIS_BUILD_DIR
 git fetch origin gh-pages:origin/gh-pages && git checkout origin/gh-pages -b gh-pages || git checkout --orphan gh-pages
--- a/.travis.d/deploy_docs.sh
+++ b/.travis.d/deploy_docs.sh
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
+set -e
+
+cd $TRAVIS_BUILD_DIR/docs && doxygen
+
 cd $TRAVIS_BUILD_DIR
-git fetch origin gh-pages:origin/gh-pages
-git checkout origin/gh-pages -b gh-pages || git checkout --orphan gh-pages
+git fetch origin gh-pages:origin/gh-pages && git checkout origin/gh-pages -b gh-pages || git checkout --orphan gh-pages
 git reset
 ls | grep -v 'docs' | xargs rm -rf
 mv docs/html/* .
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ addons:
             - doxygen
 
 script:
-    - cd $TRAVIS_BUILD_DIR/docs && doxygen
+    - echo "TODO need to build testing with samples"
 
 before_deploy:
     - cd $TRAVIS_BUILD_DIR
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,20 @@
 sudo: false
 
-addons:
-    apt:
-        packages:
-            - doxygen
+cache:
+    directories:
+        - $HOME/doxygen
+
+before_script:
+    # download doxygen 1.8.12
+    - mkdir -p $HOME/doxygen
+    - cd $HOME/doxygen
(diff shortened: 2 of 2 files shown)


d3m3vilurr:
wait.... i found mistypo

d3m3vilurr:
done.

.travis.d/deploy_docs.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+cd $TRAVIS_BUILD_DIR
+git fetch origin gh-pages:origin/gh-pages
+git checkout origin/gh-pages -b gh-pages || git checkout --orphan gh-pages
+git reset
+ls | grep -v 'docs' | xargs rm -rf
+mv docs/html/* .
+rm -rf docs
+git add *
```
devnoname120: It may be better to use `git add -A`.

.travis.yml
```diff
@@ -0,0 +1,24 @@
+sudo: false
+
+addons:
+    apt:
+        packages:
+            - doxygen
```
d3m3vilurr: actualy, docker-base travis only support 12.04. it related travis run speed.
i want to avoid version holding until found specific reason. it can make slow testing.
current released ubuntu still use v1.8.11
so if you want to hold v1.8.12, we need compile that version.

i only found one diff things, v1.7.x version not support markdown.(`support markdown` is main feature of v1.8.x)
IMO, it can acceptable. but i'll check another way.)

.travis.d/deploy_docs.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+cd $TRAVIS_BUILD_DIR
+git fetch origin gh-pages:origin/gh-pages
+git checkout origin/gh-pages -b gh-pages || git checkout --orphan gh-pages
```
d3m3vilurr: at the develop this script, actually not pushed gh-pages in repo, and i want to update or create.

well.... yeah currently it's useless and can broken in that case
first time i made `fetch && checkout -b || git checkout --orphan`
it changed after fixing, so i'll return original cases..
but IMO, still need handling not exist gh-pages case.
if break at deploy time, travis will say `build fail`

.travis.d/deploy_docs.sh
d3m3vilurr: i want to exclude hidden files(startswith . files)
`git add -A` will add all files include hidden file.

.travis.yml
devnoname120: For production, we definitely want to use a specific version. The documentation builder should not use a new version without that we manually verify that everything works, otherwise it will break when we don't expect it to.

Either upload the 1.8.12 DEB somewhere or build this version in the script.

devnoname120:
Why do you use an external script file instead of putting everything in the travis file?

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+
+script:
+    - cd $TRAVIS_BUILD_DIR/docs && doxygen
```
devnoname120: Is this pull request a WIP then?

.travis.d/deploy_docs.sh
devnoname120: OK, we want Travis to tell us it failed if the branch doesn't exist.

.travis.d/deploy_docs.sh
devnoname120: OK.

.travis.yml
d3m3vilurr: not really. but i guess yes, until fix problem about holding doxygen version and solve other review things.

.travis.d/deploy_docs.sh
d3m3vilurr: it's complicate. actually we use github environ. travis also join this system.
but, in `git` environ,  `fork repo` or `clone remote repo` not mean copied all branches.

so, in my cases, i want not receiving fail this cases, it's just repo problem, not a branches problem.
i guess, problem is travis shown build failure as broken deploy, it's false alarm.

anyway, really `we` want this, i'll fix them. :/

d3m3vilurr:
i wrote comment before, i want ci test only test pr commit things.
(so i'll make building `samples` in travis. but actually this feature posit outside in this pr :))
`deploy document` is useful feature for end users, but not related build test.
we only need to guarantee, it's latest updated version..

so i select `script` instead other types, but in my test, multiple commands [occur error](https://travis-ci.org/d3m3vilurr/vita-headers/builds/176774847)
i don't know why, but i decide to split commands to shell script at that time.
hm...

.travis.yml
devnoname120: OK, build + cache seems to be the best way.

.travis.d/deploy_docs.sh
devnoname120: Hm OK, I suppose we can keep it as is then.

devnoname120:
It seems that having an external deploy script is the norm, so you're right.

About the Travis error, does replacing `script: bash $TRAVIS_BUILD_DIR/.travis.d/deploy_docs.sh` with `script: $TRAVIS_BUILD_DIR/.travis.d/deploy_docs.sh` fix the problem?
This seems to be an argument problem so it's probably expecting a file path rather than a command.

d3m3vilurr:
Done.

- now use prebuilt doxygen v1.8.12 and store cache
- only deploy in vitasdk's repo

.travis.yml
```diff
@@ -19,6 +24,7 @@ deploy:
     skip_cleanup: true
     script: bash $TRAVIS_BUILD_DIR/.travis.d/deploy_docs.sh
     on:
+        repo: vitasdk/vita-headers
```
devnoname120: Doesn't it prevent non-vitasdk members from easily enabling documentation deployment on their fork so as to test it?

d3m3vilurr: yes, if someone want to deploy their repo, just need remove that line and push empty gh-pages branch.

[this](https://travis-ci.org/d3m3vilurr/vita-headers/builds/177312256) is deploy test in local repo
[this](https://travis-ci.org/d3m3vilurr/vita-headers/builds/177312697) is prevent deploy result.

d3m3vilurr:
@devnoname120 revert changes.
i don't want more fixing about this. do not change your mind.

devnoname120:
I changed my mind because we discussed it over IRC. What's the purpose of a discussion if one shouldn't change his mind?

PR #97, opened 2016-11-18
Add button intercept headers

yifanlu:


include/psp2/ctrl.h
```diff
@@ -257,6 +257,24 @@ int sceCtrlGetControllerPortInfo(SceCtrlPortInfo *info);
  * @return 0, <0 on error.
  */
 int sceCtrlGetBatteryInfo(int port, SceUInt8 *batt);
+
+/**
+ * Sets intercept
```
devnoname120: Could you explain more what enabling intercept does?

yifanlu: Done

PR #98, opened 2016-11-19
Renamed SCE_CTRL_ALL to SCE_CTRL_INTERCEPTED

devnoname120:
The old wording was incorrect, if we wanted a macro for masking all buttons, it should be a OR'ing of the above button macros, and it would give 0xffff.

### DIFF
--- a/include/psp2/ctrl.h
+++ b/include/psp2/ctrl.h
@@ -33,23 +33,23 @@ enum {
 /** Enumeration for the digital controller buttons.
  * L1/R1/L3/R3 only can bind using sceCtrlReadBufferPositiveExt2 */
 enum {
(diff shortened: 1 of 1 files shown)


d3m3vilurr:
yeah. it's PS button. i guess if we block PS button popup using `sceShellUtilLock`, it can read using `sceCtrlPeekBufferPositive`

devnoname120:
@TheOfficialFloW You mean that my wording is correct for user, but in kernel this value is instead used for the home button or do I misunderstand you?

d3m3vilurr:
few hours ago, I also tested this pr in user mode application
0x10000 meaning is right.(sorry, I was leaved my comment before testing)
all interrupted popup(like press ps button, holding power button) make same values.

not tested kernel plugin yet... (I'm lazy...)

Issue #80, opened 2016-11-07
Discussion: Standards for entries in db.json


devnoname120:
Currently, some function names are prefixed with a '_' or two. The exact meaning of this prefixing is still [debated](https://github.com/vitasdk/vita-headers/pull/79).

Since they only convey information that is useless for homebrew developers, should we delete them all?
From a homebrew developer point of view, adding an underscore to some functions but not to others is a annoyance.

In addition, the function names come from very different sources: it's currently inconsistent, and it would be very hard to fix the names to a chosen standard.

Do you agree with the suggestion of deleting all prefixed underscores from `db.json`?

@yifanlu, @173210, @xyzz, @DaveeFTW

173210:
There are some other symbols which have different prefixes such as `vsh`. We may make clear what prefixes are reserved by the SDK.

DaveeFTW:
Ok I've given this some thought and think I'll write this to begin some discussion by laying out some ideas. Since we've already established headers and names for user types (so far), we should exclude them from these discussions and opt in for compatibility with existing homebrew.
Kernel API is a blank canvas for us, we're still in beta(tm) and can change as much as we want. So I suggest for kernel:

- wherever possible take inspiration from user mode equivalents and suffix ```ForKernel```. e.g, ```sceIoOpen``` to ```sceIoOpenForKernel```. We would not use the ```ForDriver``` suffix.

For new contributions:
- if we have an "official" name we will use it only for reference.
- names will begin with ```sce``` prefix, followed by a reasonable* name or abbreviation for the module it belongs to.
- names will be ```camelCaseStyle```.

As such, all ```sce``` prefixes will be reserved by SDK.
\* I'm not sure what reasonable is, maybe we should create a table? The obvious exception is using ```sceKernel``` when it comes to the various ``ForKernel``` libraries.

d3m3vilurr:
how about just use `k` prefix or suffix for the kernel function?
`ForKernel` is too long suffix.

DaveeFTW:
We could do two prefixes:

- ```ksce``` for kernel functions
- ```sce``` for user functions

This would make it very clear that the developer is using the intended function.

DaveeFTW:
For the record: adding a kernel prefix is not a technical limitation but a purposeful constraint to clearly convey to the developer which library they're attempting to use. Otherwise a misguided developer may link their kernel plugin against the user stubs.

yifanlu:
[...]

PR #99, opened 2016-11-20
begin kernel renaming.

DaveeFTW:
remove all ```ForDriver``` and ```ForKernel``` suffix and prefix every kernel function with ```ksce``` instead of ```sce```. Related issue #80


xyzz:
Is it possible that there are two functions named `xxxForKernel` and `xxxForDriver` with differing implementations?

yifanlu:
@xyzz not that I've seen

devnoname120:
@DaveeFTW What's up about the renaming?

devnoname120:
`db.json` is left untouched because it will be removed anyway at some point.

include/psp2kern/bt.h
```diff
@@ -281,52 +281,52 @@ typedef struct _SceBtHidRequest {
...
-int sceBtSetL2capEchoResponseBufferForTestForDriver(int r0, int r1, int r2, int r3);
-int sceBtSetStatusForTestForDriver(int r0, int r1, int r2, int r3);
-int sceBtStartAudioForDriver(int r0, int r1, int r2, int r3);
-int sceBtStartConnectForDriver(unsigned int mac0, unsigned int mac1); // mac0 can be used as a device index (already paired devices)?
-int sceBtStartDisconnectForDriver(unsigned int mac0, unsigned int mac1);
-int sceBtStartInquiryForDriver(void);
-int sceBtStopAudioForDriver(int r0, int r1, int r2, int r3);
-int sceBtStopInquiryForDriver(void);
-int sceBtUnregisterCallbackForDriver(SceUID cb);
+int ksceBtAvrcpReadVolumeForDriver(int r0, int r1, int r2, int r3);
```
DaveeFTW: Remove ```ForDriver``` suffix

TheOfficialFloW:
I am being conservative about this and have to say that I am totally against this prefix.
I don't see any reason to have suffix or prefix that indicate kernel functions. It has besn explained to me that the reason behind this is to be able to differentiate kernel functions from the user ones and thus developers won't get confused. But why would anybody pick the wrong functions when there's a separate folder called psp2kern for kernel headers? Also every kernel library has got the suffix ForKernel/ForDriver, that's enough of indication for the developer.
As a reverse engineer I'd like to have everything as original as possible. I don't think Sony named their functions with ForDriver/ForKernel.
In my opinion we should even omit it and let's have same names for kernel and user. One could also use two different db.yml for kernel and users if it's about being able to pick correct functions from the database.

Anyways, if you guys all really want to do it, I don't want to stop you, but I agree with @yifanlu that using the k prefix for syclib is a bad idea since one would need to rename everything when using open source libraries.

DaveeFTW:
@TheOfficialFloW Can you explain what you mean by "original as possible?" and provide an example where reverse engineering is hindered by not having an "original" name? My understanding is that we have almost zero "official" kernel names and unlikely to stumble across any more in the future. Striving for official names is unrealistic.

The SDK automatically links a significant portion of user libraries to assist developers. If we match names between user and kernel then kernel plugins will likely be compiled against user libraries or fail altogether without special flags. This can be frustrating to developer that are less experienced than yourself. If we have the correct flags, it is still possible for developers to link kernel libraries against user ones manually (or vice versa). Having a very simple namespace between the two APIs provides the needed abstraction to prevent all of this without significant re-engineering. This issue has already manifested several times throughout the beta period.

For future readers: the argument for dropping the prefix on sysclib for maintaining compatibility is a compromise. Using the sysclib functions is not recommended as we cannot (have not/will not) verify whether the implementations are actually conforming to the C standard or will/wont continue to in the future. memcpy in the past did not conform to the standard for example.
If you require access to libc-like functions in your plugin opt to use code from another source such as libk - which is a port of pdclib.

TheOfficialFloW:
@DaveeFTW I mean from the "reverse engineer's" point of view, not that it is hindering reverse engineers.

devnoname120:
@TheOfficialFloW
I was first also against this change, but finally decided that it was a good idea for multiple reasons. One reason is auto-completion. Some people use IDE's, and prefixing the Kernel functions with k make it way easier to see what you can use when writing a program. Another reason is that a Kernel plugin should only contain k-prefixed functions, so this is way easier to spot incorrect uses, especially when copy-pasting code.

> Also every kernel library has got the suffix ForKernel/ForDriver, that's enough of indication for the developer.

This is not true, there are many occurrences of Kernel-only functions that don't have the suffix: `sceKernelLoadModuleForPid()`, `sceKernelRxMemcpyKernelToUserForPid()`, and many more.

TheOfficialFloW:
@devnoname120 I mean the suffix of the stubs defined in Makefile.

yifanlu:
@TheOfficialFloW so there's a couple of problems we are trying to solve with this change

1. There is an inconsistency in how stuff is previously named. Kernel functions had either no suffix, or "ForDriver" or "ForKernel" at the end. The last two scheme was done arbitrarily by Proxima and I because we saw that the libraries had that suffix. However, in the case of SceKernelThreadmgr, notice that many of the NIDs for the ForDriver and ForKernel libraries are the same indicating that their original names were the same. Having both ForDriver and ForKernel is confusing in that respect.
2. I don't like using the preposition "for" in these functions. Take the case of "sceKernelLoadModuleForPidForKernel" It indicates we are loading a module "for" another process but it is done "by" the kernel. Maybe we can do "sceKernelLoadModuleForPidByKernel" but then "sceIoOpenByKernel" sounds awkward. Having a prefix is easier to deal with in this case.

Your suggestion of just dropping prefix/suffix altogether doesn't work in all cases either. We have many user syscalls that can be called in kernel that might share the same name with a kernel function in kernel (SceCtrl is a good example, where the syscall version sees the flags differently). In these case, we might have both imports in the same application and therefore they require different names.

yifanlu:
@TheOfficialFloW we will merge this unless you have any farther objections?

TheOfficialFloW:
@yifanlu
Well ok, you guys can do it.

PR #103, opened 2016-11-27
Sce bt for driver nids

xerpi:


include/psp2kern/bt.h
```diff
@@ -0,0 +1,334 @@
+/**
```
devnoname120: Please abide by the new documentation conventions. See [this](https://github.com/vitasdk/vita-headers/blob/b3e049f48f4248272d8624501bacfbe474ffcc50/docs/definitions.h#L85-L86), and [this](https://github.com/vitasdk/vita-headers/blob/b3e049f48f4248272d8624501bacfbe474ffcc50/include/psp2/io/fcntl.h#L1-L4).

xerpi: Sure! I will send a PR soon™.

PR #105, opened 2016-11-29
Add database in yaml format

frangarcj:


db.yml
```diff
@@ -0,0 +1,5177 @@
+version: 0x2
+firmware: 3.6
+modules:
```
devnoname120: Maybe add `type: "imports"`?

PR #109, opened 2016-12-09
Add travis tests

frangarcj:


.travis.yml
```diff
@@ -12,8 +12,24 @@ before_script:
 
 script:
     - cd $TRAVIS_BUILD_DIR
-    - echo "TODO need to build testing with samples"
-
+    - curl https://api.github.com/repos/vitasdk/autobuilds/releases  | grep browser_download_url | grep linux | head -n 1 | cut -d '"' -f 4 | xargs wget
```
d3m3vilurr: why mix curl + wget?
how about change to
`curl https://api.github.com/repos/vitasdk/autobuilds/releases  | grep browser_download_url | grep linux | head -n 1 | cut -d '"' -f 4 | xargs curl -L | tar xj` ?

.travis.yml
```diff
@@ -12,8 +12,24 @@ before_script:
...
+    - cd $TRAVIS_BUILD_DIR
+    - cp -r include/* $VITASDK/arm-vita-eabi/include
+    - vita-libs-gen db.yml output
+    - cd output
+    - make
+    - cp *_stub.a $VITASDK/arm-vita-eabi/lib
+    - cd $TRAVIS_BUILD_DIR
+    - git clone https://github.com/vitasdk/samples
+    - cd samples
+    - for f in */Makefile; do make -C ${f%/*} all; done
```
d3m3vilurr: looks nice!

.travis.yml
frangarcj: I'm always using that line, no other reason

d3m3vilurr: yes. i saw before. it's just my curious thing(and also little things).
curl and wget are very similar application, so we can reduce this complex

options:
1. `curl` change to `wget -qO-`
2. `wget` change to `curl` or `curl -L`

both cases can join next tar extract command.

Issue #118, opened 2017-01-05
Please add scePower NIDs


Yoti:
`int scePowerRequestColdReset(void);`
`int scePowerRequestStandby(void);`
`int scePowerRequestSuspend(void);`
https://forums.vitasdk.org/viewtopic.php?f=4&t=14

devnoname120:
They are already in the NID database (https://github.com/vitasdk/vita-headers/blob/master/db.yml).

Feel free to add them to [`power.h`](https://github.com/vitasdk/vita-headers/blob/master/include/psp2/power.h) and open a pull request.

PR #119, opened 2017-01-05
Update power.h

DerColonel:
Added cold reset, shutdown and suspend functions as requested in #118.

`int scePowerRequestColdReset(void);`
`int scePowerRequestStandby(void);`
`int scePowerRequestSuspend(void);`

In my tests a succeeding call returned 0. I don't know in what case the call might fail and what it would return in this case.

I found functions with the same name in a PSP SDK documentation which said they always return 0. Therefore I assumed the same would be true here.

Please correct it if anyone knows better.


devnoname120:
Thanks for your contribution. 👍

PR #120, opened 2017-01-12
Added usb nids and headers

TheOfficialFloW:


devnoname120:
@d3m3vilurr Please make sure [`definitions.dox`](https://github.com/vitasdk/vita-headers/blob/master/docs/definitions.dox) is updated as well before merging.

Otherwise this happens: 
vitasdk miscategorization

d3m3vilurr:
ouch... sorry. but this file is little annoying and too often broken. how about make pre checking script for this? maybe we add to ci

devnoname120:
This would be good, but I don't feel like maintaining this script.

PR #122, opened 2017-01-15
Add helper script for definitions.dox

d3m3vilurr:
- append to travis
- fix headers & definitions.dox

related #120

### DIFF
--- a/.travis.d/definition_check.py
+++ b/.travis.d/definition_check.py
@@ -4,7 +4,8 @@ import re
 import fnmatch
 
 CURR_DIR = os.path.dirname(os.path.realpath(__file__))
-DEF_FILE_PATH = os.path.join(CURR_DIR, '..', 'docs', 'definitions.dox')
+DEF_FILE = 'definitions.dox'
+DEF_FILE_PATH = os.path.join(CURR_DIR, '..', 'docs', DEF_FILE)
 INCLUDE_DIR = os.path.join(CURR_DIR, '..', 'include')
 
 DEFINE_RULE = re.compile(r' \*     \\defgroup ([a-zA-Z]+) [a-zA-Z ]+')
@@ -46,20 +47,22 @@ def check_headers(definitions):
                 if not m:
                     continue
                 if have_group_define:
-                    errors.append('Has multiple groups: %s' % header_file)
+                    errors.append('%s: Has multiple groups' % header_file)
                     break
                 group = m.group(2)
                 if definitions.get(group) == None:
-                    errors.append('Unknown group: %s' % m.group(2))
+                    errors.append('%s: Unknown group %s' %
+                                  (header_file, m.group(2)))
                     break
                 definitions[group] += 1
                 have_group_define = True
             if not have_group_define:
-                errors.append('Could not find definition: %s' % header_file)
+                errors.append('%s: Could not find definition' % header_file)
     # reverse check if exist header
     for k, v in definitions.iteritems():
         if v == 0:
-            errors.append('Could not find using header: %s' % k)
+            errors.append('%s: Could not find using header: %s' %
+                          (DEF_FILE, k))
     return errors
 
 if __name__ == '__main__':
--- a/.travis.d/definition_check.py
+++ b/.travis.d/definition_check.py
@@ -40,17 +40,21 @@ def check_headers(definitions):
         if header_file in IGNORE_FILES:
             continue
         with open(header_path, 'r') as h:
+            have_group_define = False
             for line in h.xreadlines():
                 m = USER_GROUP_RULE.match(line)
                 if not m:
                     continue
+                if have_group_define:
+                    errors.append('Has multiple groups: %s' % header_file)
+                    break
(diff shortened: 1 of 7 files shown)


.travis.d/definition_check.py
```diff
@@ -0,0 +1,67 @@
...
+import sys
+import re
+import fnmatch
+
+CURR_DIR = os.path.dirname(os.path.realpath(__file__))
+DEF_FILE_PATH = os.path.join(CURR_DIR, '..', 'docs', 'definitions.dox')
+INCLUDE_DIR = os.path.join(CURR_DIR, '..', 'include')
+
+DEFINE_RULE = re.compile(r' \*     \\defgroup ([a-zA-Z]+) [a-zA-Z ]+')
+USER_GROUP_RULE = re.compile(r' \* \\(user|kernel)group\{([a-zA-Z]+)\}')
```
devnoname120: Do we want it to be indentation-sensitive? It will fail finding the group if there is an extra space.

.travis.d/definition_check.py
```diff
@@ -0,0 +1,67 @@
...
+            for line in h.xreadlines():
+                m = USER_GROUP_RULE.match(line)
+                if not m:
+                    continue
+                usergroup = m.group(2)
+                if definitions.get(usergroup) == None:
+                    errors.append('Unknown group: %s' % m.group(2))
+                else:
+                    definitions[usergroup] += 1
+                break
```
devnoname120: Make sure that `\(user|kernel)group{}` is present only once.

.travis.d/definition_check.py
```diff
@@ -0,0 +1,67 @@
...
+            if not m:
+                continue
+            definitions[m.group(1)] = 0
+    return definitions
+
+def check_headers(definitions):
+    errors = []
+    # check exists in definitions
+    for header_path in findfile(INCLUDE_DIR, '*.h'):
+        header_file = header_path.split('include/')[1]
```
devnoname120: Should we force headers in psp2 to use `\usergroup{}` and those in psp2kern to use `\kernelgroup{}`?

include/psp2kern/net/net.h
```diff
@@ -1,4 +1,9 @@
 /**
+ * \kernelgroup{SceNet}
+ * \usage{psp2kern/net/net.h,-lSceNetPs_stub}
```
devnoname120: `-lSceNet_stub`

.travis.d/definition_check.py
d3m3vilurr: yes. we don't need mixed indentation. all files should have same format. isn't it?

.travis.d/definition_check.py
d3m3vilurr: maybe we can check as next steps, but this script introduced for only one reason about cross checking between definitions.dox and headers.

include/psp2kern/net/net.h
d3m3vilurr: SceNetPs is right.
https://github.com/vitasdk/vita-headers/blob/11abeb6/db.yml#L3111

.travis.d/definition_check.py
d3m3vilurr: fixed. just rename to `group`

.travis.d/definition_check.py
devnoname120: Yes, but someone doing a pull request might get frustrated trying to understand why the group is not detected even though it's there.

.travis.d/definition_check.py
devnoname120: OK, that would be a good thing to do.

.travis.d/definition_check.py
devnoname120: The break statement prevents `definitions[usergroup]` from being greater than 1.
Also, no error is raised if it's greater than 1.

d3m3vilurr: no, they use both kernelgroup and usergroup. so i change variabl name to just `group`
and these group use same name. currently each definition group can be 1 or 2.

.travis.d/definition_check.py
d3m3vilurr: 1. we should detect wrong indentation.
2. but your opinion is sometime we should pass that.

i'm not agree these conflict rules.
reviewer can suggest to contributors.

.travis.d/definition_check.py
devnoname120: I have a hard time understanding you.

What I'm saying is that the following won't cause the script to return an error:
```C
/**
 * \usergroup{SceFoo}
 * \usergroup{SceBar}
 * \usage{psp2/npdrmpackage.h,-lSceNpDrmPackage_stub}
 */
```

d3m3vilurr: ah.... sorry, that's my eng problem :'(

first, i thought one header only have one user or kernel group.
so i'm not check next line and just break in loop

I need rule about that, you want to raise error if have multiple define. right? or need to check second define?

PR #127, opened 2017-02-07
Rename ksceKernelMemPool* functions

TheOfficialFloW:
Since SceUsbAudio exposes the real name of `ksceKernelMemPoolCreate` which is actually `sceKernelCreateHeap`.

### DIFF
--- a/db.yml
+++ b/db.yml
@@ -1644,10 +1644,10 @@ modules:
           ksceKernelGetMemBlockBase: 0xA841EDDA
           ksceKernelFindMemBlockByAddr: 0x8A1742F6
           ksceKernelRemapBlock: 0xDFE2C8CB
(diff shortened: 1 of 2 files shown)


d3m3vilurr:
LGTM, but this changes will break homebrew plugins include taiHEN.

https://github.com/search?q=ksceKernelMemPoolCreate&type=Code

yifanlu:
Before merging this, please submit a PR to taiHEN and you might also want to alert @xerpi too.

devnoname120:
@yifanlu Maybe add `#define ksceKernelMemPoolCreate ksceKernelCreateHeap` somewhere? Not sure if it would be possible to warn the user it's deprecated.

TheOfficialFloW:
@yifanlu @xerpi Is this ready to be merged?

PR #136, opened 2017-02-11
Added dozen of kernel nids and rearranged database

TheOfficialFloW:


devnoname120:
If possible, please split in multiple commits the new NIDs and the rearrangement.

TheOfficialFloW:
Ough, that's going to be hard and I'm also not up to doing that tbh.

db.yml
```diff
@@ -1697,7 +1715,6 @@ modules:
           memset: 0xAB9BF5C
           rshift: 0x1D89F6C0
           snprintf: 0xAE7A8981
-          sprintf: 0x3DDBE2E1
```
devnoname120: Why?

TheOfficialFloW: This NID was added by me, but I have finally realized that it is not sprintf.

devnoname120:
Human-readable diff: https://www.diffchecker.com/oJUF5DWC

devnoname120:
See https://github.com/vitasdk/vita-headers/pull/138.

PR #141, opened 2017-02-20
Added some sblauthmgr functions

yifanlu:
Added ksceKernelGetPaddr to headers

### DIFF
--- a/include/psp2kern/sblauthmgr.h
+++ b/include/psp2kern/sblauthmgr.h
@@ -1,0 +1,24 @@
+/**
+ * \kernelgroup{SceSblAuthMgr}
+ * \usage{psp2kern/sblauthmgr.h,?}
+ */
+
+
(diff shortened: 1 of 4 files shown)


include/psp2kern/sblauthmgr.h
```diff
@@ -0,0 +1,24 @@
+/**
+ * \kernelgroup{SceSblAuthMgr}
+ * \usage{psp2kern/sblauthmgr.h,?}
```
devnoname120: Replace the question mark with `-lSceSblAuthMgrForKernel_stub`.

include/psp2kern/sblauthmgr.h
```diff
@@ -0,0 +1,24 @@
+/**
+ * \kernelgroup{SceSblAuthMgr}
```
devnoname120: Add the group there as well: https://github.com/vitasdk/vita-headers/blob/master/docs/definitions.dox

PR #151, opened 2017-03-28
Fixing Mostly Harmless Typos (And One Error)

hexxellor:
This fixes a bunch of miscellaneous typos and grammatical mistakes in
the source which were mostly harmless with the exception of the
ScreenShotParam in psp2/screenshot.h which didn’t match the name of the
struct causing it to fail when using that header.

### DIFF
--- a/include/psp2/screenshot.h
+++ b/include/psp2/screenshot.h
@@ -1,5 +1,5 @@
 /**
- * \usergroup{SceScreenshot}
+ * \usergroup{SceScreenShot}
  * \usage{psp2/screenshot.h,-lSceScreenShot_stub}
  */
 
--- a/include/psp2/screenshot.h
+++ b/include/psp2/screenshot.h
@@ -1,5 +1,5 @@
 /**
- * \usergroup{SceScreenShot}
+ * \usergroup{SceScreenshot}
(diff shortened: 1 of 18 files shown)


include/psp2/screenshot.h
```diff
@@ -1,5 +1,5 @@
 /**
- * \usergroup{SceScreenShot}
+ * \usergroup{SceScreenshot}
```
devnoname120: Revert this commit and change it here too instead: https://github.com/vitasdk/vita-headers/blob/3bafb9a504a51966ff2df5a602da9775373b4409/docs/definitions.dox#L224

CelesteBlue-dev: Sorry but is that not a feet on back ? francar renamed SceScreenshot functions for a reason...

hexxellor: @CelesteBlue-dev True, but it was renamed to ScreenShot everywhere else by francar, so that should probably match too, right? 😉

devnoname120: @frangarcj capitalized the second "s" of the "Screenshot" occurrences, so the change I'm requesting here is in line with it.

devnoname120:
Thanks for your changes! 👍

CelesteBlue-dev:
Thank you both.

PR #153, opened 2017-04-01
Fix SceDriverUser libs, add SceMotionDev NIDs

xerpi:


xerpi:
Urgh yeah, gonna send a patch to the samples.

PR #158, opened 2017-04-09
Add GPIO header and more NIDs

xerpi:


include/psp2kern/lowio/gpio.h
```diff
@@ -0,0 +1,47 @@
+/**
+ * \kernelgroup{SceGpio}
+ * \usage{psp2kern/lowio/gpio.h,?}
```
devnoname120: Replace `?` with `-lSceGpioForDriver_stub`.

devnoname120:
Please add the group there as well: https://github.com/vitasdk/vita-headers/blob/master/docs/definitions.dox

docs/definitions.dox
```diff
@@ -96,6 +96,14 @@
  * \}
  *
  *
+ * \defgroup Lowio Low-level I/O
```
devnoname120: Could you move the group `Lowio` under `Filesystem`?

PR #160, opened 2017-04-10
Added missing kernel IO headers.

Rinnegatamante:
Added missing kernel IO headers.

### DIFF
--- a/include/psp2kern/io/devctl.h
+++ b/include/psp2kern/io/devctl.h
@@ -1,0 +1,39 @@
+/**
+ * \kernelgroup{SceDevCtl}
+ * \usage{psp2kern/io/devctl.h,?}
+ */
+
+
+#ifndef _PSP2_IO_DEVCTRL_H_
+#define _PSP2_IO_DEVCTRL_H_
+
+#include <psp2/types.h>
+
+#ifdef __cplusplus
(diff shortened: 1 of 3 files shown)


xerpi:
Looks good to me.

d3m3vilurr:
looks good to me either.
but just curious, anyone can explain why we broke CamelCase?
eg;
- `ksceIoDevctl` vs `ksceIoDevCtl`
- `ksceIoGetstatByFd` vs `ksceIoGetStatByFd`

xerpi:
The userspace functions are
```
sceIoDevctl
```
and
``` 
ksceIoGetstatByFd
```
so not sure if they are wrong.

TheOfficialFloW:
Sony defined them this way

include/psp2kern/io/devctl.h
```diff
@@ -0,0 +1,39 @@
+/**
+ * \kernelgroup{SceDevCtl}
+ * \usage{psp2kern/io/devctl.h,?}
```
devnoname120: Please change `?` to `-lstub_name`.
The `?` left in the include files are there out of laziness, it's not an example to follow.

Rinnegatamante:
Should be fine now.

PR #161, opened 2017-04-10
Added missing lib refs, fixed enum name in jpegenc.

Rinnegatamante:
Added missing lib refs, fixed enum name in jpegenc.

### DIFF
--- a/include/psp2/gxt.h
+++ b/include/psp2/gxt.h
@@ -1,6 +1,6 @@
 /**
  * \usergroup{SceGxt}
- * \usage{psp2/gxt.h,?}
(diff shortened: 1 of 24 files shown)


devnoname120:
`SceLibKernel_stub.a` is included by default, no need to add thé link hint.

Rinnegatamante:
So, should i remove such reference? It was already used for several headers (in the old SceKernel_stub form).

devnoname120:
Yes, it would be nice if you could remove the other locations where is appears. :+1:

Rinnegatamante:
Done

Issue #123, opened 2017-01-15
Update library-linking hints in the documentation for CMake


devnoname120:
```
\usage{psp2kern/net/net.h,-lSceNet_stub}
```

should be replaced with:

```
\usage{psp2kern/net/net.h, libSceNet_stub.a}
```

Some tutorial regarding those should be added too.

PR #162, opened 2017-04-10
Remove -l prefix from library linking hints

devnoname120:
This is long overdue.
#123


CelesteBlue-dev:
@devnoname120 Why removing -l ? And in Makefile, have us to edit lib imports or not ?

devnoname120:
The `-l` syntax was too specific.
We recommend using `CMakeLists.txt` files, plain `Makefile` should not be used.

Check the [samples](https://github.com/vitasdk/samples) to see how to use CMake.

Rinnegatamante:
psp2/io/fcntl.h requires Sceiofilemgr_stub (https://github.com/vitasdk/vita-headers/blob/master/db.yml#L2143) but in current header file is reported as it doesn't require a linked library.

devnoname120:
Please make a new pull request to address this issue.

PR #163, opened 2017-04-16
Fix some incorrect module and library NIDs

### LINKED ISSUE #7: Use real library/module NIDs in included db
In the included databases, it seems like the module/library nids all have placeholder values. For extras, we should either come up with a NID generation system (ideally, reverse sony's linker and see how nids are generated), which is needed for vitasdk to support creating dynamic libraries or use hard coded values for now.

For the sony imports, the module nids never change, so we should keep them in the database. Library nids do change, so we can either remove that field from the specifications or mark it optional (and make it an array of "known nids" or something). The advantage for the second option is that it may be useful in the future.

devnoname120:
Fix #7

### DIFF
--- a/db.yml
+++ b/db.yml
@@ -2,7 +2,7 @@ version: 2
 firmware: 3.60
 modules:
   SceAppMgr:
-    nid: 0x0
+    nid: 0xDBB29DB7
     libraries:
       SceAppMgr:
         functions:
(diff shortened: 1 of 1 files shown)


devnoname120:
Some modules such as `SceAudioIn` don't actually exist. We have made the decision on IRC to keep those anyway to avoid breaking the compilation of existing homebrews.

devnoname120:
For the record, this was done by doing [`./dbtools.py fixnids`](https://github.com/devnoname120/vitasdk-db-tools).

PR #165, opened 2017-04-21
Added vitasdk.h equivalent for kernel.

Rinnegatamante:
Added vitasdk.h equivalent for kernel.


CelesteBlue-dev:
@Rinnegatamante you are right. Why not call it kernel.h like on official SDK ?

Rinnegatamante:
Coherency with psp2kern nomenclature.

CelesteBlue-dev:
ok.

PR #167, opened 2017-05-05
Update usbd.h

yifanlu:
Argument 1 was wrong


xerpi:
I didn't do a very meticulous RE so so yeah, that arg could be wrong.

xyzz:
I didn't review it but it's by yifan so must be correct.

PR #168, opened 2017-05-11
Fixed a couple function descriptions and added a few missing functions in ScePower

joel16:
Title explains it all. I've tested these functions in VITAident, and I can assure you they do indeed work correctly. 

Reference: https://github.com/joel16/VITAident/blob/master/src/power.c

Please let me know if you require any changes..

### DIFF
--- a/include/psp2/power.h
+++ b/include/psp2/power.h
@@ -84,9 +84,9 @@ SceBool scePowerIsPowerOnline(void);
 int scePowerGetBatteryLifeTime(void);
 
 /**
- * Returns battery remaining capacity ?
+ * Returns battery remaining capacity
  *
- * @return ?
+ * @return battery remaining capacity in mAh (milliampere hour)
  */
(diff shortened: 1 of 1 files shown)


include/psp2/power.h
```diff
@@ -98,13 +98,41 @@ int scePowerGetBatteryRemainCapacity(void);
...
+ * Returns battery temperature
+ *
+ * @return temperature in degrees celcius * 100
+ */
+int scePowerGetBatteryTemp(void);
+
+/**
+ * Returns battery voltage
+ *
+ * @return battery voltage in kilo volts
```
d3m3vilurr: It looks milivolt, isn't it?

joel16: My bad, you're correct. I had it the wrong way round. 
Fixed with https://github.com/vitasdk/vita-headers/pull/168/commits/779a39336345281208280e5c68220979787863a1

PR #169, opened 2017-05-11
Add sceRegMgr functions

joel16:
I'm not exactly sure where this would fit in the docs -> definitions.dox, so please let me know and I'll make the changes.

### DIFF
--- a/docs/definitions.dox
+++ b/docs/definitions.dox
@@ -230,7 +230,10 @@
  *      Export a photo from a file or data
  *
  *     \defgroup ScePss PSS Code Memory Handling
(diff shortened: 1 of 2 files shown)


devnoname120:
You need to define the group in `docs/definitions.dox` as well.

joel16:
I'm aware of that, but what I was asking is I'm not sure what grouping you'd want me to put this under. Example, defgroup Miscellaneous or defgroup SystemStatus?

devnoname120:
Maybe `SystemParams`?

joel16:
Done, thanks for looking into it noname sensei.

PR #170, opened 2017-05-12
Added some ScePaf functions

TheOfficialFloW:


db.yml
```diff
@@ -5724,3 +5724,17 @@ modules:
...
           ksceMotionDevSetSamplingMode: 0xDBAF611A
           ksceMotionDevSamplingStop: 0xFD0B0785
+  ScePaf:
+    nid: 0xCD679177
+    libraries:
+      ScePafStdc:
+        kernel: false
+        nid: 0xA7D28DAE
+        functions:
+          sce_paf_private_free: 0x1B77082E
```
d3m3vilurr: can you change these functions like `scePafPrivateFree`?

TheOfficialFloW: They were named this way on PSP and I guess the name has been inherited.

d3m3vilurr: 1. have only one reason for the compatibility between psp sdk?
2. we don't make header file for these nids at name time?

currently, I'm ok, but it looks breaking of convention, so I want to another reviewer's opinions

ps. maybe I'm wrong, we just decided `_` prefix, not decided camelSase and snake_case.. #80

d3m3vilurr:
```
17:34:02 F<frangarcj> PSS functions don't follow convention too
```

PR #175, opened 2017-05-14
Update vitasdk.h with recent headers

joel16:


devnoname120:
Please update `vitasdkkern.h` as well.

joel16:
I noticed psp2kern/bt.h, and psp2kern/lowio/gpio.h were left in out in the header. Was this intentional? Just want to make sure before I added those in.

devnoname120:
I think that you can add them.

Issue #110, opened 2016-12-09
Add names to anonymous enumerations


devnoname120:
Example in display.h:

```C
 enum {
     SCE_DISPLAY_SETBUF_IMMEDIATE = 0,
     SCE_DISPLAY_SETBUF_NEXTFRAME = 1
 };
```

should be replaced with

```C
typedef enum SceDisplaySetBufSync {
     SCE_DISPLAY_SETBUF_IMMEDIATE = 0,
     SCE_DISPLAY_SETBUF_NEXTFRAME = 1
 } SceDisplaySetBufSync;
```

It can then be referenced in the documentation as `::SceDisplaySetBufSync`.

d3m3vilurr:
all enum and defined type start upper case. so should be `SceDisplaySetBufSync`

devnoname120:
Thanks, I modified the description accordingly.

Rinnegatamante:
Can be closed i suppose.

PR #176, opened 2017-05-16
Added sceKernelPowerLock sceKernelPowerUnlock

yifanlu:


devnoname120:
The enum that is referred with `::KernelPowerTickType` is actually anonymous, this should be changed (#110).
Since it's already referenced by other parts of this file, I'll merge the pull request anyway.

PR #177, opened 2017-05-18
Fix ksceNetSend() and ksceNetRecv()

devnoname120:
These functions don't exist in Kernel. #define have been added to make them work.

@Cpasjuste Please review. :)

### DIFF
--- a/include/psp2kern/net/net.h
+++ b/include/psp2kern/net/net.h
@@ -706,9 +706,7 @@ int ksceNetSocket(const char *name, int domain, int type, int protocol);
 int ksceNetAccept(int s, SceNetSockaddr *addr, unsigned int *addrlen);
 int ksceNetBind(int s, const SceNetSockaddr *addr, unsigned int addrlen);
 int ksceNetListen(int s, int backlog);
-int ksceNetRecv(int s, void *buf, unsigned int len, int flags); //!< This function does not actually exist in Kernel, so it's a #define to ksceNetRecvfrom() for compatibility
 int ksceNetRecvfrom(int s, void *buf, unsigned int len, int flags, SceNetSockaddr *from, unsigned int *fromlen);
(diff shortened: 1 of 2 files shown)


d3m3vilurr:
why not use `#define`?

include/psp2kern/net/net.h
```diff
@@ -706,10 +706,14 @@ int ksceNetSocket(const char *name, int domain, int type, int protocol);
 int ksceNetAccept(int s, SceNetSockaddr *addr, unsigned int *addrlen);
 int ksceNetBind(int s, const SceNetSockaddr *addr, unsigned int addrlen);
 int ksceNetListen(int s, int backlog);
-int ksceNetRecv(int s, void *buf, unsigned int len, int flags);
+int ksceNetRecv(int s, void *buf, unsigned int len, int flags); //!< This function does not actually exist in Kernel, so it's a #define to ksceNetRecvfrom() for compatibility
 int ksceNetRecvfrom(int s, void *buf, unsigned int len, int flags, SceNetSockaddr *from, unsigned int *fromlen);
-int ksceNetSend(int s, const void *msg, unsigned int len, int flags);
+int ksceNetSend(int s, const void *msg, unsigned int len, int flags); //!< This function does not actually exist in Kernel, so it's a #define to ksceNetSendto() for compatibility
```
d3m3vilurr: ok, my comment was wrong. but if just remove this prototype, also not need db.yml changes.
1. why add db.yml?
2. why add `#define`

its looks duplicated.

devnoname120: 1. I removed it from `db.yml` because it was incorrect.
2. `#define` is the right way to go. I added the function prototype as well so that it appears in the documentation with the rest of the functions.

d3m3vilurr: ah, my bad... I saw with a unified view...
so, we haven't `ksceNetSend`. why provide this function information?
and doxygen will add `#define` info..

devnoname120: The `#define` section is further down, so I thought that it would be more clear for doc readers to leave the function prototype. Do you want me to remove it?

PR #178, opened 2017-05-18
Properly named anonymous enums.

Rinnegatamante:


Rinnegatamante:
I'm working on naming every single anonymous enum in the headers. I'll make a new pull request when ready.

PR #182, opened 2017-05-26
Add SceHid Mouse functions

meetpatty:
Add new SceHid Mouse functions and fix up some SceHidKeyboard definitions.


devnoname120:
@meetpatty Is it WIP or ready for merging?

PR #183, opened 2017-06-05
Removed no more used db.json.

Rinnegatamante:


devnoname120:
Note: https://github.com/173210/vita-analyze still uses it.

Rinnegatamante:
I don't see why we should still have an heavly outdated db.json.
If it's used somewhere, it should be updated, if not, it can be removed.

devnoname120:
Let's get rid of it. If someone still needs it, it's easy to write a tool to produce it from`db.yml`.

PR #188, opened 2017-06-15
Added missing stuffs to jpegenc.h.

Rinnegatamante:


Rinnegatamante:
@xerpi proposed some changes on #henkaku for this.

PR #190, opened 2017-06-17
Renamed enums values, typo fix in jpegenc.h.

Rinnegatamante:


xerpi:
I hope the hb breakage is minimal.

PR #191, opened 2017-06-17
Added kernel mount/unmount headers

yifanlu:


include/psp2kern/io/fcntl.h
```diff
@@ -318,6 +318,30 @@ int ksceIoChangeAsyncPriority(SceUID fd, int pri);
...
+  * Unmounts a device
+  *
+  * @param[in]  id    Device to unmount
+  * @param[in]  a2    
+  * @param[in]  a3    
+  * @param[in]  a4    
+  *
+  * @return < 0 on error.
+  */
+int ksceIoUmount(int id, int a2, int a3, int a4);
```
devnoname120: Please provide a known-working value for the unknown parameters.

yifanlu: I've only seen zeros. You can add that but idk much about this function.

yifanlu:
Don't merge, there's a problem.

PR #197, opened 2017-06-22
change SceDebugForDriver printf naming and add definition

Cpasjuste:
I think the function name may be wrong, as it may be difficult, sometime, to redefine printf..
This, at least, allow kernel debug print from [psp2shell](https://github.com/Cpasjuste/psp2shell).

### DIFF
--- a/include/psp2kern/kernel/sysmem.h
+++ b/include/psp2kern/kernel/sysmem.h
@@ -203,6 +203,8 @@ int ksceKernelGetPaddr(void *addr, uintptr_t *paddr);
 
 int ksceSysrootIsManufacturingMode(void);
 
+int ksceDebugPrintf(const char *fmt, ...);
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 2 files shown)


include/psp2kern/kernel/sysmem.h
```diff
@@ -203,6 +203,10 @@ int ksceKernelGetPaddr(void *addr, uintptr_t *paddr);
 
 int ksceSysrootIsManufacturingMode(void);
 
+int ksceDebugPrintf(const char *fmt, ...);
+
+int ksceDebugPrintf2(int unk0, int unk1, const char *fmt, ...);
```
devnoname120: Please add a line of doc where you give known working valies for the unk arguments. This saves frustration from developers willing to use this function.

d3m3vilurr: I think she just got a NID info & reverse info, and only use dumping log from kernel
https://github.com/Cpasjuste/psp2shell/blob/master/psp2shell_k/main.c#L82-L86
so I though, she can't know mean of args.

Cpasjuste:
Yes, unfortunately i don't know the two first args :/

PR #200, opened 2017-06-23
Check dup function definition and missing nid

d3m3vilurr:
I'm not sure, this regex rule is correct or not.
But probably help to fix NID mismatch.


d3m3vilurr:
merged 745de8d58e0df6e1ba57335f2be3c4276315ae53

PR #208, opened 2017-06-30
Fix errors on sceCamera enums / sceKernelGetMutexInfo params

codestation:


devnoname120:
These structures are also referenced in some comments in the same file. Could you fix them as well?

codestation:
@devnoname120 i checked the comments on camera.h and don't see anything out of place with the comments referencing SceCameraDevice / SceCameraPriority / SceCameraFrameRate / SceCameraExposureCompensation. There is something i am missing?

devnoname120:
My mistake, I thought that I saw some incorrect references but there aren't.

PR #211, opened 2017-07-21
Added PSBUTTON alias for SCE_CTRL_INTERCEPTED, added a note for shell privileges request.

Rinnegatamante:


joel16:
What about SCE_CTRL_HOME, just for consistency purposes since none of the other enums are named with  'button' in the end?

Rinnegatamante:
It's called "PS Button" even in SceShellSvc: https://github.com/vitasdk/vita-headers/blob/master/include/psp2/shellutil.h#L15

PR #214, opened 2017-09-03
Added more sysmodule definitions

TheOfficialFloW:


TheOfficialFloW:
I have made a HENkaku PR for this change: https://github.com/henkaku/henkaku/pull/84

TheOfficialFloW:
only those that are using the package installer need to be updated with the new names, and there are just VitaShell and HENkaku

devnoname120:
There is also VHBB but I will update it.

PR #216, opened 2017-09-10
Use real names of SysEvent

TheOfficialFloW:
The original name of the nid is found in SceHid.


d3m3vilurr:
sadly this patch will break xyzz/gamecard-microsd ...

PR #219, opened 2017-09-23
Add some sce_paf_private functions

devnoname120:


devnoname120:
`sce_paf_private_strncpy: 0x5883A6E3` looks wrong, it's actually `sce_paf_private_memcpy`.

TheOfficialFloW:
Can you test it?

TheOfficialFloW:
Also why are there two memmove? Please note the difference in the header

devnoname120:
The assembly of `0x5883A6E3` directly calls `sceClibMemcpy`, and it doesn't set the last byte to `\0`.

devnoname120:
The two memmove have some slightly different assembly but they do the exact same thing

TheOfficialFloW:
Oh ok, feel free to remove it then.

PR #223, opened 2017-09-29
Update db.yml (SceAppMgr add a lot of NIDs) CB#1

CelesteBlue-dev:


db.yml
```diff
@@ -33,13 +61,13 @@ modules:
...
           _sceAppMgrGetAppMgrState: 0x55C60D20
-          _sceAppMgrGetAppParam: 0x258C3E3C
           _sceAppMgrGetAppParam2: 0x26D43923
+          _sceAppMgrGetAppParam: 0x258C3E3C
           _sceAppMgrGetBootParam: 0xEEFBC756
           _sceAppMgrGetBudgetInfo: 0xEACF2237
           _sceAppMgrGetCoredumpStateForShell: 0x3DD7E65
-          _sceAppMgrGetCurrentBgmState: 0x81569A58
           _sceAppMgrGetCurrentBgmState2: 0x3F7B9BB5
+          _sceAppMgrGetCurrentBgmState: 0x81569A58
```
devnoname120: Please revert this change and the other changes where `function2` was moved before `function`.

CelesteBlue-dev:
@devnoname120 "Please don't remove the leading 0s from hexadecimal numbers, they should always be in the form of 0xXXXXXXXX."
Do you mean that I have to keep 0x01234567 instead of writing 0x1234567 ?

d3m3vilurr:
```
> git grep ": 0x[0-9a-fA-F]\{1,7\}\$" db.yml | grep sce | wc -l
184
```
hm... :/

Issue #201, opened 2017-06-23
NID mismatch / duplicate function definition


d3m3vilurr:
#200 found these
```
psp2kern/bt.h: Could not find NID ksceBtPairingOOB
psp2kern/udcd.h: Could not find NID ksceUdcdGetDrvList
psp2kern/udcd.h: Could not find NID ksceUdcdWaitCancel
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelChangeCurrentThreadAttr
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelReleaseWaitThread
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelGetThreadId
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelGetThreadExitStatus
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelCheckThreadStack
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelGetThreadInfo
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelGetThreadRunStatus
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelWaitSemaCB
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelCancelSema
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelGetSemaInfo
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelOpenMutex
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelCloseMutex
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelLockMutexCB
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelGetMutexInfo
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelGetEventFlagInfo
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelGetCallbackInfo
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelCreateMsgPipe
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelDeleteMsgPipe
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelSendMsgPipe
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelSendMsgPipeCB
psp2kern/kernel/threadmgr.h: Could not find NID ksceKernelTrySendMsgPipe
[...]

PR #229, opened 2017-10-07
Rename sceCompatGetDevInf to sceCompatGetDevInfo

d3m3vilurr:
related #201


TheOfficialFloW:
Change the header file instead. We should keep the original name

d3m3vilurr:
related #80
I guess, already we changed some user function name. is that rule only impact kernel nid?

PR #244, opened 2017-10-28
Various fix/update to modulemgr.h

yne:
Fix invalid module_stop definition
Add missing module_exit
Change module_* to function pointer

### DIFF
--- a/include/psp2/kernel/modulemgr.h
+++ b/include/psp2/kernel/modulemgr.h
@@ -27,8 +27,7 @@ extern "C" {
 #define SCE_KERNEL_STOP_CANCEL        SCE_KERNEL_STOP_FAIL
 /** @} */
 
-typedef struct SceKernelSegmentInfo
-{
+typedef struct SceKernelSegmentInfo {
 	SceUInt size;   //!< sizeof(SceKernelSegmentInfo)
 	SceUInt perms;  //!< probably rwx in low bits
 	void *vaddr;    //!< address in memory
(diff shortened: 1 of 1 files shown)


include/psp2/kernel/modulemgr.h
```diff
@@ -37,23 +37,34 @@ typedef struct SceKernelSegmentInfo
 	SceUInt res;    //!< unused?
 } SceKernelSegmentInfo;
 
+typedef struct {
```
xerpi: I think the convention is to put the struct name after "typedef struct" too.

include/psp2/kernel/modulemgr.h
```diff
@@ -37,23 +37,34 @@ typedef struct SceKernelSegmentInfo
...
+	SceUInt32 ptr;  //!< prel31 offset to the start of the function
+	SceUInt32 insn; //!< start of the function (80 B0 B0 B0), 1=cannot be unwound
+} SceKernelModuleExidx;
+
+typedef struct {
```
xerpi: Same here, struct name missing.

include/psp2/kernel/modulemgr.h
```diff
@@ -37,23 +37,34 @@ typedef struct SceKernelSegmentInfo
...
+	SceUInt32 insn0;  //!< 
+	SceUInt32 insn1;  //!< 
+	SceUInt32 flags;  //!< often 0
+} SceKernelModuleExtab;
+
 typedef struct
 {
```
xerpi: Mind placing the `{` bracket on the same line as typedef struct and also adding the name?

include/psp2/kernel/modulemgr.h
yne: `SceKernelLMOption`, `SceKernelULMOption` and `SceKernelModuleInfo` does not follow theses rules neither.
Should I also update them ?

include/psp2/kernel/modulemgr.h
yne: `SceKernelSegmentInfo` and `SceKernelModuleInfo` does not follow these rules.
Should I also update them ?

Rinnegatamante:
Where's the missing module_exit added?

yne:
in the staging area, on my laptop, from two years ago.
By the time I get home and `grep -r` my (probably removed) vitasdk folder, you would had time to create a PR ;)

PR #255, opened 2017-11-29
Add psp2kern/{lowio/pervasive.h, power.h, syscon.h, uart.h}, minor nits

xerpi:


docs/definitions.dox
```diff
@@ -258,6 +264,14 @@
  * \}
  *
  *
+ * \defgroup Syscon Syscon
+ *  Syscon processor communication
+ * \{
+ *     \defgroup SceSyscon Syscon communication Library
```
devnoname120: Maybe put it in the `Lowio Low-level I/O` group instead? I would prefer to avoid to create a new root group just for Syscon.

PR #264, opened 2017-12-20
Added some missing stuffs in sceCtrl header.

Rinnegatamante:


include/psp2/ctrl.h
```diff
@@ -82,14 +87,32 @@ typedef struct SceCtrlData {
...
+	/** Down button */
+	uint8_t		down;
+	/** Left button */
+	uint8_t		left;
 	/** Left trigger (L2) */
-	unsigned char	lt;
+	uint8_t		l2;
 	/** Right trigger (R2) */
-	unsigned char	rt;
+	uint8_t		r2;
```
d3m3vilurr: it can make compatibility issue.
can you remain lt and rt?

Rinnegatamante: I think it would be better to name it r2/l2 due to the fact we have also l1/t2 references.
Afaik, the only plugins using these are @xerpi ones to adapt DS3/DS4 to Vita via BT.

d3m3vilurr: personally, I like l2/r2 than lt/rt too. but the library should have low broken changes. (although that was bad decision)
these fields can use all homebrews not only use plugins. (for example moonlight used [these][1])

[1]: https://github.com/xyzz/vita-moonlight/blob/8c71034/src/input/vita.c#L454-L455

include/psp2/ctrl.h
```diff
@@ -82,14 +87,32 @@ typedef struct SceCtrlData {
...
+	/** Left button (L1) */
+	uint8_t		l1;
+	/** Right button (R1) */
```
d3m3vilurr: these comments should change to just `L1 button` or something.

Rinnegatamante: I named them in that way cause in modern Dualshocks, L1 is the equivalent of LB and L2 the equivalent of LT in x360 joystick (so Left Button/Left Trigger).

PR #267, opened 2017-12-27
db.yml major update

Rinnegatamante:
Major update for db.yml:

- Added several missing modules.
- Added several missing functions.
- Replaced some fake module nids with correct ones.
- Moved kernel and nid parameters before functions one in modules where they were swapped so that we're now using the same order for all modules.
- Added missing 0 to some functions that were using nids with <8 digits.


d3m3vilurr:
Squash and merged 958313b018fbb6be0

PR #277, opened 2017-12-31
Fix travis errors

d3m3vilurr:
- `libSceCompat_stub` make userland library
- hide make stdout; it looks annoying
- change language type to c for `cmake`

this patch needs vitasdk/samples#41

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ script:
     - cp -r include/* $VITASDK/arm-vita-eabi/include
     - vita-libs-gen db.yml output
     - cd output
-    - make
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -25,12 +26,16 @@ script:
     - cp -r include/* $VITASDK/arm-vita-eabi/include
     - vita-libs-gen db.yml output
     - cd output
-    - make
+    - make > /dev/null
```
devnoname120: 👍

yne: what is the difference with `make -s` ?

PR #286, opened 2018-01-16
Added sceFpu header.

Rinnegatamante:


include/psp2/fpu.h
```diff
@@ -0,0 +1,147 @@
+/**
+ * \usergroup{SceFpu}
+ * \usage{psp2/fpu.h,SceFpu_stub}
+ */
```
d3m3vilurr: @Rinnegatamante already you said on the IRC, this module cannot use this lib for the hbl develop like other static libs.
can you add this information to here?

Rinnegatamante:
Where should i add such info?

d3m3vilurr:
use bottom the `\usage`. that's okay :>

PR #288, opened 2018-01-27
Added missing definitions for VitaShell and adhoc headers

TheOfficialFloW:


include/psp2/pspnet_adhocctl.h
```diff
@@ -0,0 +1,82 @@
+/**
+ * \usergroup{ScePspnetAdhocctl}
```
devnoname120: Needs to be defined in `definitions.dox`.

devnoname120:
Merged.

PR #295, opened 2018-02-15
Removing static libs nids.

Rinnegatamante:


CelesteBlue-dev:
Why are you removing some functions ?

yne:
@CelesteBlue-dev  those functions are not part of the Vita OS libraries.
They are only useful if you have the source to compile those sce* libraries, and even so, it would make more sense to have they export in the corresponding {{module}}.yaml.

Having static/non-core libs exports listed in the db.yaml is stupid.

PR #300, opened 2018-03-05
Added db for 3.65/3.67

TheOfficialFloW:
We need to have a way to specify 3.65/3.67 NIDs. Maybe use `*_367_stub.a`?

### DIFF
--- a/db_367.yml
+++ b/db_367.yml
@@ -1,0 +1,86 @@
+version: 2
+firmware: 3.67
+modules:
+  SceSysmem:
+    nid: 0x3380B323
+    libraries:
+      SceSysmemForKernel:
+        kernel: true
+        nid: 0x02451F0F
+        functions:
+          ksceKernelCreateUidObj: 0xFB6390CE
+          ksceKernelRxMemcpyKernelToUserForPid: 0x2995558D
+          ksceKernelGetMemBlockType: 0xD44FE44B
+          ksceKernelFindClassByName: 0x7D87F706
+      SceCpuForKernel:
+        kernel: true
+        nid: 0xA5195D20
+        functions:
+          ksceKernelCpuDcacheInvalidateAll: 0xF9B0B171
+          ksceKernelCpuDcacheWritebackAll: 0x49180814
+          ksceKernelCpuDcacheWritebackInvalidateAll: 0x4BBA5C82
+          ksceKernelCpuIcacheInvalidateRange: 0x2E637B1D
+          ksceKernelCpuIcacheInvalidateAll: 0x803C84BF
+          ksceKernelCpuIcacheAndL2WritebackInvalidateRange: 0x73E895EA
+          ksceKernelCpuDcacheWritebackInvalidateRange: 0x4F442396
+      SceDebugForKernel:
+        kernel: true
+        nid: 0x13D793B7
+        functions:
+          ksceDebugSetHandlers: 0x88AD6D0C
+          ksceDebugRegisterPutcharHandler: 0x22546577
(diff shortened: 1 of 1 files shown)


CelesteBlue-dev:
the NIDs of 3.60 functions have changed on 3.65/3.67 or are only some new functions ?

xerpi:
An improved proposal based on yours:
Let's imagine `foo_stub.a` exports the function `bar`.
By using weak symbols we make `foo_368_stub.a` export `bar` as a weak symbol alias and `bar_368` regularly so one can link with both `foo_stub.a` and `foo_368_stub.a`, and if you call `bar` it will call the regular version, while both version of the function will be callable at runtime.
So in `foo_368_stub.a`, `bar` would need to be a weak alias to `bar_368`.

CelesteBlue-dev:
@xerpi do you mean you are proposing to call ksceMemcpy_368(); when compiling for 3.68 ?

xerpi:
@CelesteBlue-dev that would be necessary only if you want to call the 368 version of `ksceMemcpy` and you are linking with *both* stubs.
If you only link with `*_368_stub.a`, thanks to the alias, calling `ksceMemcpy` would also be possible.

CelesteBlue-dev:
@xerpi I think the easiest way is to edit vita-elf-create or such to add --version argument.
When -v 368, it will link sceMemcpy_stub_368.a.
By default, to keep compatibility, when version is not specified it will link 360 stubs.

xerpi:
@CelesteBlue-dev but you don't even need to touch vita-elf-create for that, only your Makefile.

CelesteBlue-dev:
yes but with my way, you would only have to do make 368 or make 365 or make (by default 360).
Now I see what you mean : 
```C
if (is_368())
     func_368()
else
     func()
```
but so which libs do you import ? 360 stubs or 368 stubs ?

xerpi:
My 'way' isn't different than yours but only an improvement that also allows you to be able to choose what function to call at runtime.

CelesteBlue-dev:
@xerpi sorry I don't understand how a skprx can be loaded if it imports 368 and 360 stubs.
it will forcibly return SCE_KERNEL_ERROR_NO_LIB...

xerpi:
Hmm so if an SKPRX imports an invalid library it returns an error when loading it? I thought it would just leave the stubs uninitialized.

TheOfficialFloW:
I don't think it will work either, since you'll end up with two libraries with the same name but different library nid. Maybe it works if we declare it as weak import, can somebody test?

CelesteBlue-dev:
@xerpi I tried loading a skprx that imported 360 NIDs on my 365 PSVita using your kernel loader and it returns an error : 
SCE_KERNEL_ERROR_MODULEMGR_NO_LIB | 0x8002d003 | C1-2712-2
or maybe also  : 
SCE_KERNEL_ERROR_MODULEMGR_INVALID_LIB | 0x8002d009 | C1-2718-8
SCE_KERNEL_ERROR_MODULEMGR_INVALID_STUB | 0x8002d00a | C1-2719-9
SCE_KERNEL_ERROR_MODULEMGR_NO_FUNC_NID | 0x8002d00b | C1-2720-1
SCE_KERNEL_ERROR_MODULEMGR_NO_VAR_NID | 0x8002d00c | C1-2721-2

xerpi:
Yeah maybe weak imports might do the trick. My thought was that the Vita's dynamic loader would skip unknown libraries intead of erroring.

d3m3vilurr:
replaced #592

PR #302, opened 2018-03-15
 Add _sceSblACMgrIsGameProgram in db.yml and headers

CelesteBlue-dev:


include/psp2/sblacmgr.h
```diff
@@ -0,0 +1,22 @@
...
+#ifndef _PSP2_SBLACMGR_H_
+#define _PSP2_SBLACMGR_H_
+
+#include <psp2/types.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int _sceSblACMgrIsGameProgram(void);
```
d3m3vilurr: can you remove underscore?

CelesteBlue-dev: no because sceSblACMgrIsGameProgram (without underscore) is a wrapper to this function.

PR #304, opened 2018-03-18
add the content_id param index.

SilicaAndPina:


Rinnegatamante:
These are just param values from param.sfo, no need to list all possible values.

PR #310, opened 2018-04-08
Add ScePafMisc lib and sce_paf_misc_does_file_exist

SilicaAndPina:
sce_paf_misc_does_file_exist(char *path);
returns 1 if it exists
0 if it does not

### DIFF
--- a/db.yml
+++ b/db.yml
@@ -8044,7 +8044,7 @@ modules:
           sce_paf_private_strrchr: 0xFA2C892F
       ScePafMisc:
         kernel: false
-        nid: 0x3d643ce8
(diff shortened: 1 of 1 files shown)


db.yml
```diff
@@ -8042,6 +8042,11 @@ modules:
           sce_paf_private_strncasecmp: 0xA6014289
           sce_paf_private_strncmp: 0x1E1EA818
           sce_paf_private_strrchr: 0xFA2C892F
+      ScePafMisc:
+        kernel: false
+        nid: 0x3d643ce8
```
Rinnegatamante: Uppercase

PR #314, opened 2018-04-21
Workaround for non 64-bit systems

frangarcj:


d3m3vilurr:
@frangarcj can you explain this patch?

d3m3vilurr:
@frangarcj said this patch is hack for the vita3k.

PR #329, opened 2018-06-29
Fix 0x50FE3B4D from ksceSysrootIsManufacturingMode to ksceSysrootUseI…

CelesteBlue-dev:
…nternalStorage

Thanks to SKGleba

### DIFF
--- a/db.yml
+++ b/db.yml
@@ -2058,7 +2058,7 @@ modules:
           ksceSysrootIsAuCodecIcConexant: 0x46E72428
           ksceSysrootIsBsodReboot: 0x4373AC96
           ksceSysrootIsExternalBootMode: 0x89D19090
-          ksceSysrootIsManufacturingMode: 0x55392965
+          ksceSysrootUseExternalStorage: 0x55392965
           ksceSysrootIsSafeMode: 0x834439A7
           ksceSysrootIsUpdateMode: 0xB0E1FC67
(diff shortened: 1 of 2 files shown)


d3m3vilurr:
1.5yr ago, this nid was added from @TheOfficialFloW #139 then removed from @yifanlu cuz dup nid #172 
should review from one of them.

yifanlu:
Please don’t remove headers. This isn’t how sdks work.

CelesteBlue-dev:
@d3m3vilurr I have already explained to yifan why this is not IsManufacturingMode function and why it is UseExternalStorage.
The IsManufacturingMode function exists under this form only:
```c
int sceSblPmMgrGetCurrentMode(int *mode);
```
Returns 0 on success. mode is set to 1 if PSVita is in manufacturing mode.

yifanlu:
Okay cool so we’re breaking homebrew code again?

CelesteBlue-dev:
nothing uses this function for now as it was not useful to know if vita was in manufacturing mode.
Now that we know it indicates if memcard is allowed to be used, it will maybe be used in storage plugins...

d3m3vilurr:
@yifanlu I said before about the version, but nobody start this :( anyway, maybe we need define macro for avoid sdk break

d3m3vilurr:
@yifanlu pushed 0ebd19f0c8bdb then made issue #331

yifanlu:
Thanks!

PR #334, opened 2018-07-17
Add sceSblSsUpdateMgrGetSpkgInfo

CelesteBlue-dev:
From https://github.com/henkaku/henkaku/blob/master/plugin/kernel.c#L230


d3m3vilurr:
official nid is probably `sceSblUsGetSpkgInfo`. but already we used `sceSblSsUpdateMgr*` instead `sceSblUs*`;
eg. `sceSblSsUpdateMgrVerifyPupHeader` vs `sceSblUsVerifyPupHeader`

PR #335, opened 2018-07-18
Fix ksceSblACMgrIsUpdate

CelesteBlue-dev:
Fix commit #2e2b259

### DIFF
--- a/db.yml
+++ b/db.yml
@@ -7733,7 +7733,11 @@ modules:
           ksceSblACMgrIsNonGameProgram: 0x6C5AB07F
           ksceSblACMgrIsDevelopmentMode: 0xE87D1777
           ksceSblACMgrIsPspEmu: 0xFD00C72A
-          ksceSblACMgrIsUpdate: 0x930CD037
+      SceSblACMgrForKernel:
+        kernel: true
+        nid: 0x11F9B314
+        functions:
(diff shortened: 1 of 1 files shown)


db.yml
```diff
@@ -7733,7 +7733,11 @@ modules:
           ksceSblACMgrIsNonGameProgram: 0x6C5AB07F
           ksceSblACMgrIsDevelopmentMode: 0xE87D1777
           ksceSblACMgrIsPspEmu: 0xFD00C72A
-          ksceSblACMgrIsUpdate: 0x930CD037
+      SceSblACMgrForKernel:
+        kernel: true
+        nid: 0x11F9B314
+        functions:
+          ksceSblACMgrIsShellForKernel: 0x930CD037
```
d3m3vilurr: in before we decided to remove all `ForKernel` and `ForDriver` from function name, and to change `sce` prefix to `ksce`.
that's a reason why we have a diff name between official sdk & nid

so, now we have a problem, `ksceSblACMgrIsShell` is already used in `SceSblACMgrForDriver`, and probably you think `ksceSblACMgrIsUpdate` is a wrong name.
some cases have `2` postfix(but I'm not like this. anyway).
are you have any suggest another name?

TheOfficialFloW:
Just remove it from the db :)

d3m3vilurr:
done 7ac549e78

PR #344, opened 2018-08-19
Added sceAppUtilAppEventParseLiveArea prototype.

Rinnegatamante:


include/psp2/apputil.h
```diff
@@ -158,6 +158,9 @@ int sceAppUtilShutdown();
 //! Receive app event
 int sceAppUtilReceiveAppEvent(SceAppUtilAppEventParam *eventParam);
 
+//! Parse received app event from LiveArea
+int sceAppUtilAppEventParseLiveArea(SceAppUtilAppEventParam *eventParam, char *buffer);
```
DaveeFTW: Is const-ness correct here?

devnoname120: [This](https://github.com/Rinnegatamante/vitaQuake/blob/82af45c89559e48491791c1368855f17e146ec57/source/sys_psp2.c#L502) seems to indicate it's not `const`.

DaveeFTW: for SceAppUtilAppEventParam

Rinnegatamante:
Fixed and rebased

PR #348, opened 2018-08-28
Added some functions of SceKernelUtilsForDriver

Princess-of-Sleeping:


include/psp2kern/kernel/sysmem.h
```diff
@@ -463,6 +463,59 @@ int ksceDebugPutchar(int character);
 
 int ksceDebugDisableInfoDump(int flag);
 
+#define SCE_SHA1_BLOCK_SIZE		64
```
Rinnegatamante: Can you ident these using spaces instead of tabs?

Princess-of-Sleeping:
I replaced tabs with spaces.

CelesteBlue-dev:
tabs are prefered I guess

Rinnegatamante:
I was referring only to the defines, structs are better with tabs.

Princess-of-Sleeping:
How about this?

d3m3vilurr:
@Princess-of-Sleeping not need adding tab between keyward and variable name of the struct. for example `psp2/apputil.h`
also not needs adding space between define macro name and value. for example `psp2/audiodec.h`

Princess-of-Sleeping:
Please tell me if there is a problem

PR #349, opened 2018-08-29
Add psp2kern/kernel/utils.h

Princess-of-Sleeping:


d3m3vilurr:
you need to change `docs/definitions.dox` :)

PR #353, opened 2018-09-23
Add ksceGzipDecompress

Princess-of-Sleeping:


yne:
Are you sure the last argument is not some kind of `int* crc32` ?

Princess-of-Sleeping:
Yes, I fixed it because the last argument was crc32.

yne:
I just realized that `uint32_t *` may have been more appropriate.
my bad

PR #355, opened 2018-10-11
nid is wrong

KuromeSan:


yne:
Library nid also seems incorrect

KuromeSan:
There. corrected both library and module nids.

PR #356, opened 2018-10-14
Update pss.h

KuromeSan:
Add some functions

### DIFF
--- a/include/psp2/pss.h
+++ b/include/psp2/pss.h
@@ -12,7 +12,17 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+  
+typedef struct PSM_handle {
+uint32_t unk0;
+uint32_t unk1;
+uint32_t filesz;
+uint32_t unk3;
 
+} PSM_handle;
+
+int pss_crypto_open_p(PSM_handle *handle, char *path);
+char *pss_crypto_read_p(PSM_handle *handle, int ctx)
 void *pss_code_mem_alloc(SceSize *);
 void pss_code_mem_flush_icache(const void *, SceSize);
 void pss_code_mem_lock(void);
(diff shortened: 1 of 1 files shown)


include/psp2/pss.h
```diff
@@ -12,7 +12,17 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+  
+typedef struct PSM_handle {
```
xerpi: To be consistent, I'd change the name to something like `ScePssHandle`, or maybe `ScePssCryptoHandle` if this handle is only used for PSS crypto functions.

include/psp2/pss.h
```diff
@@ -12,7 +12,17 @@
...
+uint32_t unk0;
```
xerpi: Also indent the struct members (tab).

KuromeSan:
that good?

include/psp2/pss.h
```diff
@@ -12,7 +12,16 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
-
+  
+typedef struct ScePssHandle {
```
devnoname120: `ScePssHandle` is unnecessary, remove it.

include/psp2/pss.h
```diff
@@ -12,7 +12,16 @@
...
+  uint32_t unk0;
+  uint32_t unk1;
+  uint32_t filesz;
+  uint32_t unk3;
+} ScePssCryptoHandle;
+
+int pss_crypto_open_p(ScePssCryptoHandle *handle, char *path);
+char *pss_crypto_read_p(ScePssCryptoHandle *handle, int unk);
```
devnoname120: Can you provide a known good value for `unk`?

Check other source files to see how you can document a function.

KuromeSan: It's a pointer to an int appears to always be 0x1 from what I've seen ?

xerpi:
Looks good to me.

PR #358, opened 2018-10-28
add ksceKernelGetProcessMainThread

DaveeFTW:
ksceKernelGetProcessMainThread is a kernel service used to query the main thread id for a given process.


d3m3vilurr:
39bcb9db4f14

PR #359, opened 2018-10-28
add ksceKernelGetProcessStatus

DaveeFTW:
This kernel service queries the status of a given process. I'm unsure what exactly the bits represent. It seems 0x10 is related to suspension status, but I'm not confident in that assumption.


d3m3vilurr:
87879bebfdc2

PR #371, opened 2018-12-14
Update vita-headers

Princess-of-Sleeping:


Princess-of-Sleeping:
ksceZlibIsValid was my mistake

yne:
Let's hope the reviewer won't forget to squash.
@Princess-of-Sleeping If you need to add a new modification into the previous commit (in order to avoid 7 duplicate commits like `Update modulemgr.h`)
you can use :
```git commit -a --amend -C HEAD```

Princess-of-Sleeping:
I do not know what the git command is, so I will send you a pull request again

PR #376, opened 2018-12-18
Update SceNetPs

CelesteBlue-dev:


yne:
This is a breaking change, since you removed `ksceNetSocketClose`  and `sceNetConnectForSyscalls`  (but are still documented in the net.h headers).

Princess-of-Sleeping:
If it can not be compiled, developers can fix the source

CelesteBlue-dev:
fixed

CelesteBlue-dev:
fixed

yne:
Thanks, one last thing, could you do it in this order :
```
int ksceNetSocketClose(int s);
...
#define ksceNetClose ksceNetSocketClose
```
this way, all NIDS you've just added are real prototypes,
and the `#define` are only for legacy purpose.

PR #379, opened 2018-12-24
Sort db.yml

yne:
Ordering the current db.yml is the first step for a continus validation.
Then, the ordering can be checked by travis for future PR using :
```bash
LIBSEP="        functions:" # name of the yaml block we want to check
IFS=@
# iterate over each block and `sort --check` they content
for lib in $(grep "$LIBSEP\|          " db.yml | sed "s/$LIBSEP/$IFS/"); do
  printf "$lib" | grep -o "          [^:]*"| LC_COLLATE=C sort -c
done
```
I'm still unsure where to put this validation code
@CelesteBlue-dev does it seems enough ?

### DIFF
--- a/db.yml
+++ b/db.yml
@@ -105,10 +105,10 @@ modules:
           _sceAppMgrInstallDirMount: 0xED37D810
           _sceAppMgrIsCameraActive: 0x7F0C9410
           _sceAppMgrLaunchAppByName: 0xDAC0AF12
-          _sceAppMgrLaunchAppByNameForShell: 0x4C3B5BB8
           _sceAppMgrLaunchAppByName2: 0xAEF06983
           _sceAppMgrLaunchAppByName2ForShell: 0xC6BA9596
           _sceAppMgrLaunchAppByName2ndStage: 0xDEDD2AED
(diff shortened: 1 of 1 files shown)


yne:
`yq` surely is more change resistant, but all the other db.yml CI checks I found were also this *hacky*, see https://github.com/vitasdk/vita-headers/blob/master/.travis.d/definition_check.py#L51
While writing the validator I hoped that the YAML structure would not change anytime soon, then a realized that if this append, rewriting the validator would be the least of our worries :)
Maybe I should merge the sort validator into this python script ?

devnoname120:
Ah yes it's pretty bad especially since Python has good yaml support.

I think a separate script is better for separation of concerns.

d3m3vilurr:
a subject could be split to this change and CI part.
so I'll merge this pr after this comment.

when the begining time that makes the definition cheking script, vitasdk members want to avoid introduce `pyyaml` module for that.
actually, they didn't want to increase `extra dependency`, so I made that without the module.

if we can allow increase dependency, it could be changed to many things for linting.

PR #380, opened 2018-12-25
Fix SceSdif according to wiki

CelesteBlue-dev:
infos from motoharu

PR #381, opened 2018-12-26
Add db.yml ordering check in CI

yne:
Following https://github.com/vitasdk/vita-headers/pull/379#issuecomment-449836370 recommandation, I added the db ordering check in a separate travis.d sh file.
So if this thing works find, this pending PR shall fail: https://github.com/vitasdk/vita-headers/pull/380


d3m3vilurr:
LGTM

PR #393, opened 2019-02-14
db: Rename hash functions

d3m3vilurr:
all applied camelCase
it also fixed typo `ksceSblDmac5AEsCbcDec`

### DIFF
--- a/db.yml
+++ b/db.yml
@@ -8528,7 +8528,7 @@ modules:
           ksceSblAimgrGetPscode: 0xE0DC2587
           ksceSblAimgrGetPscode2: 0x9A9676D0
           ksceSblAimgrGetVisibleId: 0x04843835
-          ksceSblDmac5AEsCbcDec: 0x121FA69F
+          ksceSblDmac5AesCbcDec: 0x121FA69F
           ksceSblDmac5AesCbcDecNP: 0x1901CB5E
           ksceSblDmac5AesCbcEnc: 0xE6E1AD15
           ksceSblDmac5AesCbcEncNP: 0x711C057A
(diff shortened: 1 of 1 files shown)


CelesteBlue-dev:
nice but if you rename then better to use proper names. look wiki: WithKeygen -> NP

d3m3vilurr:
well...WithKeygen would be okay, imo that was called with `sceSblDmac5EncDecKeyGen` data by user side
but actually i don't care both, just I need definition of `NP`

CelesteBlue-dev:
NP = NoPortability. That's because these NP functions dont use enc_dec_with_portability_sm.self (a f00d module) that stores the keys.

d3m3vilurr:
ok. it reasonable name. i'll change it.

d3m3vilurr:
hm we already have `WithPortability` postfix

1. remove them
2. NP replace to `WithPortability` postfix.

what do you prefer?

CelesteBlue-dev:
@devnoname120 and me made a tool to get names from the wiki and then insert into db.yml
I think it could be easier to use this tool.
In all cases, take the names from wiki then fix typos.

d3m3vilurr:
as you knows we(vitasdk team) don't care original sce names.
we just focus that names have readability, right meaning and backward compatibility.
also this patchset is one of polishing definition after importing.

d3m3vilurr:
okay no more conversation here.
only remain issues are
1. `WithPortability` part
2. `NoPortability`(`WithKeygen`) part.

I agree that `WithKeygen` is meanless name, but `NP` also too short/meanless postfix.
just it's better than `WithKeygen`.

my options are
1. remove `WithPortability` postifx
2. we use `WithoutProtability` instead `NP`

I just merge current code and I'll make an another issue for that.

PR #401, opened 2019-03-18
libkernel variable exports

frangarcj:

PR #403, opened 2019-03-20
Rewrite ordering script

d3m3vilurr:
this patch will resolve CI breaking in #401


d3m3vilurr:
just choice #404 instead this.

Issue #331, opened 2018-07-01
need release versioning


d3m3vilurr:
we already suffered sdk break as multiple times. each time someone like me suggest versioning, then return history.

pros:
- all devs no more getting problem breaking changes until sdk version up. it mean passing role to devs from sdk team.
- can have time for refactoring(like rename) and experimental feature (like #300)
- we can collect changes list before release time.
- ~~no more get yifan's rage~~

cons:
- most cases we always used latest dev version like other platform sdks.
- who decide release & manage new version

yne:
From someone having to deal with Angular constant breaking changes, I couldn't agree more. :+1: 

> - who decide release & manage new version

I've serious doubts that someone would be delighted to do this. Futhermore, human arbitrary naming may cause drama. So the only sane solution is to define a versioning rule that everyone would agree on. 
For example, you suggested a `major.minor` naming which mean (if not manually incremented by a human) that we shall bump the major on every API breaking change, like https://github.com/vitasdk/vita-headers/pull/354 or https://github.com/vitasdk/vita-headers/pull/355.
I'm not a soothsayer but judging by the look of the accepted MR, we will catch up with Chrome version by the end of next year.
An other human-less alternative versioning could be `year.month` (a la Windows / Ubuntu), this way the increment could be done heartlessly by the CI.

> - most cases we always used latest dev version like other platform sdks.

So maybe we should live with this behavior and carry it by explicit any API breaking changes in the doxygen, so after an SDK update, developers trying to google why the function disappeared will find the new name with an explanation "Was renamed from $oldname, see #987". This shall reduce the frustration while porting to the new version. But this also may encourage to keep using master instead of version. Making all the versioning efforts worthless.

yne:
From someone having to deal with Angular constant breaking changes, I couldn't agree more.

> - who decide release & manage new version

I've serious doubts that someone would be delighted to do this. Furthermore, human arbitrary naming may cause drama. So the only sane solution is to define a versioning rule that everyone would agree on. 
For example, you suggested a `major.minor` naming which mean (if not manually incremented by a human) that we shall bump the major on every API breaking change, like https://github.com/vitasdk/vita-headers/pull/354 or https://github.com/vitasdk/vita-headers/pull/355.
I'm not a soothsayer but judging by the look of the accepted MR, we will catch up with Chrome version by the end of next year.
An other human-less alternative versioning could be `year.month` (a la Windows / Ubuntu), this way the increment could be done heartlessly by the CI.

[...]

PR #410, opened 2019-07-03
Fixed some function names and sorted alphabetically

Princess-of-Sleeping:


Princess-of-Sleeping:
I do not know this error

d3m3vilurr:
because `P` is lower value than `T`
`ksceSblAimgrIsTest` and `ksceSblAimgrIsPrototypeRev2`

db.yml
```diff
@@ -8997,972 +8754,1281 @@ modules:
...
+          sceUsbdUnregisterLdd: 0x473692CC
+      SceUsbdForDriver:
         kernel: true
-        nid: 0xB706084A
+        nid: 0xA0EBCA41
         functions:
-          ksceMsifDisableSlowMode: 0x75848756
-          ksceMsifEnableSlowMode: 0x4B751CE6
-          ksceMsifGetMsInfo: 0xD0307849
-
```
d3m3vilurr: these are original function before the patch.
so can you change other side?
and maybe give postfix(like 2~4) is better than prefix(`_`)

but if you think it's private function, just remain the comment :)

Princess-of-Sleeping: I'm sorry but I can not figure out where to fix it.
Function names such as _sceErrorGetExternalString are defined in SceSysLibTrace.

d3m3vilurr: i mentioned c2a2ec7205 commit parts, and you added same name function to other side.
imo, `_` prefix only can have an internal function(private method)
btw technically we can call it if that was defined `nid`. :/

CelesteBlue-dev: No @d3m3vilurr there are _sce prefixed functions that are exported.
The issue is the ksce naming: it would give k_sce.
Moreover sometimes __sce.

d3m3vilurr: @CelesteBlue-dev I know. and you told repeat and repeat about objection of the vitasdk's decisions that why we use `ksce` instead `ForDriver/Kernel` and why we don't use the sony's export name.
please stop.

d3m3vilurr: so @CelesteBlue-dev you can suggest another name, and if that name has really good, we can change every name.
but it is also sdk, so full changing wouldn't be acceptable :)

d3m3vilurr:
this pr would break backward compatibility, but it seems move to right place.
some modules(SceRTC, SceDisplay, SceMotion) located wrong place for passing the travis tests.
it could be next steps (related #331)

and I'll merge it manually, pr needs squashing some commits(`Fixed incorrect blahblah` and `Return Sce* blahblah`) and I don't agree to rename function of `SceSblSsUpdateMgr`

d3m3vilurr:
merged 8fe91a2

yne:
I was worried to see a PR so huge (+7,555 −7,552) being merged

PR #411, opened 2019-07-11
add kscePowerSetCallbackState

KuromeSan:


db.yml
```diff
@@ -7228,6 +7228,7 @@ modules:
           kscePowerSetArmClockFrequency: 0x74DB5AE5
           kscePowerSetBatteryFakeStatus: 0x0C6973B8
           kscePowerSetBusClockFrequency: 0xB8D7B3FB
+          kscePowerSetCallbackEnabled: 0xA6FF5997
```
d3m3vilurr: thank you @KuromeSan.
but I'm not sure that name is a good.
I couldn't expect NID function result (i'm not reversing it yet), it seems ambiguous.
it would be `setting callback feature is enabled` or `callback has been set and so it is enabled`.
how about give more good name?

KuromeSan: well when you call it with (0) argument it seems to disable the callbacks for the ps button and power button, - cant open power menu, and pressing it just turns off the screen without entering standby mode.. also (-1) enables it again . 

i cant really think of a better name since it literally enables/disables callbacks

maybe kscePowerToggleCallback?

d3m3vilurr:
merged 5df8fabd82300

PR #415, opened 2019-08-04
Partial revert  PR410

d3m3vilurr:
after merge it, we have broken changes that some NID moves to `SceDriverUser`
this patch revert this changes.
and we need to check what fields & group moves.

please show this link.
https://github.com/d3m3vilurr/vita-headers/compare/recheck-sort..partial-revert-410
it compare before #410 + apply sort and after this patch.


d3m3vilurr:
@devnoname120 if you want to revert `SceKernelThreadMgr` part, please tell me :)

d3m3vilurr:
because these nids aren't kernel functions. @Princess-of-Sleeping would be right.

devnoname120:
Are you sure that they are not kernel?
https://gist.github.com/devnoname120/2975ef91ed5180fe8cf680424ebbde66#file-db_lookup-yml-L8027-L8035

d3m3vilurr:
well, I saw https://wiki.henkaku.xyz/vita/SceDriverUser#SceErrorUser and https://wiki.henkaku.xyz/vita/SceDriverUser#SceDrmBridgeUser

and might be your gist file would be wrong. for example `SceRtcUser` marked `kernel: true`. are you sure it? our yaml has it with `kernel: false` :)

devnoname120:
I think my lookup is incorrect. Iirc I used some heuristics to determine whether it's kernel or not because the underlying tool that I used didn't provide this information.

PR #427, opened 2019-10-03
Added some function and description

Princess-of-Sleeping:


include/psp2kern/kernel/modulemgr.h
```diff
@@ -201,18 +200,307 @@ SceUID ksceKernelGetProcessMainModule(SceUID pid);
...
+ *
+ * @param[in]  pid         - target pid
+ * @param[in]  module_addr - target module address
+ * @param[out] a3          - unknown output
+ * @param[out] a4          - unknown output
+ * @param[out] module_name - module name
+ *
+ * @return 0 on success, < 0 on error.
+ */
+int ksceKernelGetModuleInfoMinByAddr(SceUID pid, const void *module_addr, int *a3, int *a4, SceKernelModuleName *module_name);
```
d3m3vilurr: how about changing the name to `ksceKerenlGetModuleNameByAddr` if this only contained the `name`?
or how about giving another struct name like `SceKernelModuleInfo3` :)?

Princess-of-Sleeping: need to test ksceKernelGetModuleInfoMinByAddr

PR #432, opened 2019-11-28
Add SceLibMtp

KuromeSan:
add SceLibMtp nids found within SceShell- noted on: https://forum.devchroma.nl/index.php/topic,53.msg66.html#msg66


db.yml
```diff
@@ -3936,7 +3936,7 @@ modules:
           pthread_vita_tls_get_np: 0xD479D238
           pthread_vita_tls_set_np: 0x483C96B2
           sched_yield: 0x473829B5
-          unlink: 0x7BBCA340
+          unlink: 0x7BBCA340
```
d3m3vilurr: could you remove these tail whitespaces?

yne:
Also why the `sceMtpBeginWhatever` namming ? is it based the official naming in somehow ? does it represent an "async" function ? In this case why not using the `sceMtpWhateverAsync` namming (just like the `sceIo` library)

KuromeSan:
The names where found within SceShell.self 
So basically it's what sony calls em

PR #438, opened 2020-01-08
Compiler-independent struct align macro

VelocityRa:
Improves compilation status on MSVC.

### DIFF
--- a/include/vitasdk/align.h
+++ b/include/vitasdk/align.h
@@ -1,0 +1,19 @@
+#ifndef _VITASDKALIGN_H_
+#define _VITASDKALIGN_H_
+
+#ifdef  __cplusplus
+extern "C" {
+#endif
+
+#if defined(_MSC_VER)
+#define SCE_ALIGN(x) __declspec(align(x))
+#else
+#if defined(__GNUC__)
+#define SCE_ALIGN(x) __attribute__ ((aligned(x)))
+#endif
+#endif
+
+#ifdef __cplusplus
+}
+#endif
+#endif
--- a/.travis.d/definition_check.py
+++ b/.travis.d/definition_check.py
@@ -27,6 +27,7 @@ FUNCTION_RULE = re.compile(FUNC_RULE_PATTERN)
 IGNORE_FILES = [
     'vitasdk.h',
     'vitasdkkern.h',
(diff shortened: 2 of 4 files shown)


include/vitasdk/align.h
```diff
@@ -0,0 +1,48 @@
...
+
+#ifdef _MSC_VER
+#    if (_MSC_VER >= 1800)
+#        define __alignas_is_defined 1
+#    endif
+#    if (_MSC_VER >= 1900)
+#        define __alignof_is_defined 1
+#    endif
+#else
+#    include <cstdalign>   // __alignas/of_is_defined directly from the implementation
```
d3m3vilurr: is it allow in gcc?

d3m3vilurr:
you should update `.travis.d/definition_check.py` to pass the travis teesting. :)

VelocityRa:
@d3m3vilurr I added it here https://github.com/vitasdk/vita-headers/pull/438/files#diff-1bf3e94b741b213928567cb72bdea7c0R30 what else should I do?

include/vitasdk/align.h
VelocityRa: Apparently your gcc version doesn't support it? Not sure, I'll remove that.

VelocityRa:
Ok builds now, used another method.

include/vitasdk/align.h
d3m3vilurr: cstdalign is c++ header. so i asked it. tha t would be only allowed with c++ flags :)

_[FastHub-Libre](https://github.com/thermatk/FastHub-Libre/)을 사용하여 Mi 5에서 보냈습니다_

VelocityRa: Ah right.

include/vitasdk/align.h
```diff
@@ -0,0 +1,19 @@
+#ifndef _VITASDKALIGN_H_
+#define _VITASDKALIGN_H_
+
+#ifdef  __cplusplus
+extern "C" {
+#endif
+
+#if defined(_MSC_VER)
+#define ALIGN(x) __declspec(align(x))
```
d3m3vilurr: could you give another name like `SCE_ALIGN` or `PSP2_ALIGN`?

PR #451, opened 2020-06-26
Fix many args

Princess-of-Sleeping:


frangarcj:
Thanks @Princess-of-Sleeping 

Next time @d3m3vilurr squash instead of merge

d3m3vilurr:
these commits was squashed :)

PR #452, opened 2020-06-27
Unify SDK Version constant for all occurrences.

Rinnegatamante:


d3m3vilurr:
@Rinnegatamante could you fix sample first? might be last vita2d patch would break the sdl2 sample.

Rinnegatamante:
If the problem is about some undefined reference to SceSharedFb functions, it's case -lSceAppMgr_stub needs to be added to the libraries list.

joel16:
Should be fixed with https://github.com/xerpi/libvita2d/pull/81#event-3489388796.

include/psp2/types.h
```diff
@@ -7,6 +7,8 @@
 #ifndef _PSP2_TYPES_H_
 #define _PSP2_TYPES_H_
 
+#define PSP2_SDK_VERSION 0x03570011
```
d3m3vilurr, line 10: how about just make the another header like `def.h`?

Rinnegatamante: With a single define?

d3m3vilurr: yep. imo, a version couldn't be types.

PR #454, opened 2020-07-08
Added Json Library. Redo

bythos14:
Added most of the exported SceLibJson functions. A couple were problematic so I omitted them. Thanks to cuevavirus for error codes.

### DIFF
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -400,7 +400,7 @@ SUBGROUPING            = YES
 # SEPARATE_MEMBER_PAGES.
 # The default value is: NO.
 
-INLINE_GROUPED_CLASSES = NO
+INLINE_GROUPED_CLASSES = YES
 
 # When the INLINE_SIMPLE_STRUCTS tag is set to YES, structs, classes, and unions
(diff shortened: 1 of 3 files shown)


d3m3vilurr:
it seems copied from official sdk. we can't allow it.
if you had reversed it, please share the information. (and at that cases, it would be c mapping)

bythos14:
What information would be sufficient and suitable? I got the function and class names from db.yml and popped them in a demangler

d3m3vilurr:
imo, you might copied error codes from official sdk (or cuevavirus did it). it's not allowed for preventing license violation.
you should remove copied parts of the pr.
(maybe we already have the copied parts in this project, but it doesn't mean, we allow to copy from leaked/official sdk)

frangarcj:
https://pastebin.com/RmfPVDs3

bythos14:
I did as you recommended for the error codes. Also made other minor changes.

d3m3vilurr:
@bythos14 if you add new changes, you should make the new pr that this pr was closed & it doesn't update anymore until reopen.

bythos14:
Okay, thank you

PR #462, opened 2020-07-24
Add sceKernelWaitSignal and sceKernelSendSignal

sunho:
While I was emulating libult in Vita3k, I found these functions are used internally to implement many sync data structures.

### DIFF
--- a/include/psp2/kernel/threadmgr.h
+++ b/include/psp2/kernel/threadmgr.h
@@ -1135,25 +1135,18 @@ int sceKernelDeleteLwCond(SceKernelLwCondWork *pWork);
 int sceKernelSignalLwCond(SceKernelLwCondWork *pWork);
 int sceKernelWaitLwCond(SceKernelLwCondWork *pWork,  unsigned int *pTimeout);
 
-typedef struct SceKernelWaitSignalResult {
-    SceUID thid;
-    SceUInt32 dret;
-}SceKernelWaitSignalResult;
-
 typedef struct SceKernelWaitSignalOptParam {
-    SceUInt32 unk[2];
-    SceKernelWaitSignalResult* result;
-}SceKernelWaitSignalOptParam;
+    SceUInt32 unk;
+} SceKernelWaitSignalOptParam;
 
 /**
  * Sleep current thread and wait for a signal. After it receives a signal, the thread wakes up.
- *
- * If it was sucessful and params is not null, params.result->dret will be 0.
+ * This is like a semphore with limit 1. If signal was sent 
  * 
  * @param params - extra parameters
  * @return 0 on success
  */
-int sceKernelWaitSignal(SceUInt32 unk0, SceUInt32 unk1, SceUInt32 unk2, SceKernelWaitSignalOptParam *params);
+int sceKernelWaitSignal(SceUInt32 unk0, SceUInt32 unk1, SceKernelWaitSignalOptParam *params);
 
 /**
  * Send a signal to another thread specified by thid.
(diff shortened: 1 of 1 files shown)


sunho:
Not fully reversed, but I think it's better than nothing...

include/psp2/kernel/threadmgr.h
```diff
@@ -1135,6 +1135,34 @@ int sceKernelDeleteLwCond(SceKernelLwCondWork *pWork);
 int sceKernelSignalLwCond(SceKernelLwCondWork *pWork);
 int sceKernelWaitLwCond(SceKernelLwCondWork *pWork,  unsigned int *pTimeout);
 
+typedef struct SceKernelWaitSignalResult {
```
Rinnegatamante, line 1138: Where is this struct used?

sunho: Right below in SceKernelWaitSignalOptParam.
It was used like below in sceult.
```c
SceKernelWaitSignalResult res;
SceKernelWaitSignalOptParam param;
param.result = &res;
sceKernelWaitSignal(0, 0, 0, &param); 
if (res.dret != 0) {
   // fail
}
```

d3m3vilurr: an another question, why return field has `d`?

sunho: I don't know. It's from sceult assert message. (The message was "dret == 0" iirc) Maybe it's unix convention. I'm seeing some unix people naming error variable as dret.

d3m3vilurr: imo, it wouldn't be unix convention. just meaning of the decimal or something.
but i see that.

include/psp2/kernel/threadmgr.h
```diff
@@ -1135,6 +1135,34 @@ int sceKernelDeleteLwCond(SceKernelLwCondWork *pWork);
...
+    SceUID thid;
+    SceUInt32 dret;
+}SceKernelWaitSignalResult;
```
d3m3vilurr, line 1141: could you add space after `}`?

include/psp2/kernel/threadmgr.h
sunho: I looked into this again. It turned out the fourth argument doesn't exist. It was just a random coincidence that an important struct used by libult was inside fourth register and not cleared.

sunho:
I updated header to be more correct.

The below sample program demonstrates the behavior of signal api.

```c
#include <psp2/kernel/threadmgr.h>
#include <psp2/kernel/processmgr.h>
#include <stdio.h>


int threadFunc(SceSize args, void *argp) {
    printf("Sleep waiter thread\n");
    sceKernelDelayThread(1*1000000); 
    printf("First wait start\n");
    sceKernelWaitSignal(0,0,0);
    printf("First wait success\n");
    printf("Second wait start\n");
    sceKernelWaitSignal(0,0,0);
    printf("Second wait success\n");
}

int main(int argc, char *argv[]) {
    int thid = sceKernelCreateThread("my_thread", threadFunc, 0x10000100, 0x10000, 0, 0, NULL);
    if (thid >= 0)
        sceKernelStartThread(thid, 0, 0);
    printf("First send\n");
    sceKernelSendSignal(thid);
    printf("Sleep sender thread\n");
    sceKernelDelayThread(3*1000000);
    printf("Second send\n");
    sceKernelSendSignal(thid);
    printf("Third send\n");
    int ret = sceKernelSendSignal(thid);
    printf("Third send error %d \n", ret);
    printf("Fourth send\n");
    ret = sceKernelSendSignal(thid);
    printf("Fourth send error %d \n", ret);
    sceKernelExitProcess(0);
    return 0;
}
```

Here's the output.

```
First send
Sleep waiter thread
Sleep sender thread
First wait start
First wait success
Second wait start
Second send
Third send
Second wait success
Third send error 0
Fourth send
Fourth send error -2147319519
```

You can see that signal is behaving like a semaphore with limit 1.

PR #463, opened 2020-07-27
Final Additions and Changes for the Json Library

bythos14:
Switched ValueType to an enum class. Changed values to be more C++ like, rather than C like.

Added public type specifier to the InitParameter class.

Removed Initializer::setAllocatorInfoCallBack and AllocatorInfoCallback typedef.

Added some missing documentation, and some other minor changes.

### DIFF
--- a/include/psp2/json.h
+++ b/include/psp2/json.h
@@ -20,18 +20,18 @@ namespace Json
 {
 
 /**
- * Enumerator of different types of Values
+ * Strongly typed Enumerator of different types of Values.
  */
-typedef enum ValueType 
+typedef enum class ValueType 
 {
-	SCE_JSON_VALUE_NULL     = 0, //!< Empty value
-	SCE_JSON_VALUE_BOOL     = 1, //!< SceBoolean value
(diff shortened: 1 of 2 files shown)


d3m3vilurr:
please recheck the indentations. they mixed soft-hard tab

bythos14:
Fixed the indentations.

include/psp2/json.h
```diff
@@ -973,7 +1050,7 @@ class Parser
 	 * 					  current character.
 	 * @param[in]  data - Data to pass to the callback function
 	 */
-	static int parse(Value& val, ParseCallback cb, void* data);
+	static int parse(Value& val, ParseCallBack cb, void* data);
```
d3m3vilurr, line 1053: I preper `Callback` than `CallBack`; now a day, it's an one word.
so how about change the `db.yml`?

bythos14: Made the change in db.yml.

d3m3vilurr:
LGTM :+1: thank you

PR #467, opened 2020-09-13
fix name of module "SceHttp"

KuromeSan:
its "SceLibHttp" ....


d3m3vilurr:
sadly, it could break many homebrew applications includes henkaku.
should we take this risk?

PR #475, opened 2020-09-22
Added error code for SceAppMgr

bythos14:
Credit also to @MerLev

Thanks to @Princess-of-Sleeping for guiding the blind.

### DIFF
--- a/include/psp2/appmgr.h
+++ b/include/psp2/appmgr.h
@@ -21,6 +21,7 @@ typedef enum SceAppMgrErrorCode {
 	SCE_APPMGR_ERROR_INVALID            = 0x8080201A, //!< Invalid param
 	SCE_APPMGR_ERROR_TOO_LONG_ARGV      = 0x8080201D, //!< argv is too long
 	SCE_APPMGR_ERROR_INVALID_SELF_PATH  = 0x8080201E, //!< Invalid SELF path
+	SCE_APPMGR_ERROR_NOEXEC             = 0x8080201F, //!< The process is not authorized to run this function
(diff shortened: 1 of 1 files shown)


d3m3vilurr:
@bythos14 could you give reference link?

bythos14:
No reference, @MERLev and I reversed that error code while trying to figure out why he couldn't load an app using SceAppMgr from the kernel.

bythos14:
There are numerous checks within SceAppMgr to ensure that only SceShell can run certain loader functions, if one of those checks fail then `0x8080201F` is returned. Two of those checks is to use ksceSblACMgrIsShell, and to check the the current process ID to see if it matches SceShell's. That's my explanation of the error code.

Princess-of-Sleeping:
`0x8080201F` is not SCE_APPMGR_ERROR_NOT_SHELL.

it SCE_APPMGR_ERROR_NOEXEC.

bythos14:
Came up with the name myself, so not surprised that it's wrong. So does this error indicate no permissions to start a separate program? Am I wrong in my assumption of it's purpose?

Princess-of-Sleeping:
In this case, it does not apply to SCE_APPMGR_ERROR_NOT_SHELL.

```
int SceAppMgrForDriver_29421DCD(int a1, ...){

	if(ksceKernelGetProcessId() != 0x10005)
		return 0x8080201F;

	/*
	 * doing some
	 */

	return res;
}
```

bythos14:
So the error will be returned when an unauthorized process tries to access functionality not allowed to it. I'll make the necessary adjustments tomorrow.

bythos14:
Adjustments made.

PR #480, opened 2020-09-25
Adjust parameters for callbacks in some libraries, and some other misc. changes.

bythos14:
Added missing param to ScePowerCallback.

Adjusted params for callbacks and fixed some indentations in SceLibJson.

Fixed a couple typos from 8ebd0b29 in fcntl.h.

Renamed `common` argument of SceKernelCallbackFunction to `userData`. Defined unknown `arg` argument in sceKernelCreateCallback as `userData`.

### DIFF
--- a/include/psp2/json.h
+++ b/include/psp2/json.h
@@ -3,7 +3,7 @@
  * \usage{psp2/json.h,SceLibJson_stub,SCE_SYSMODULE_JSON}
  * 
  * \note Compiling and linking with this library requires you 
- * 		 to compile your code with the compiler flag -fno-rtti.
+ *       to compile your code with the compiler flag -fno-rtti.
  */
 
 #ifndef _PSP2_JSON_H_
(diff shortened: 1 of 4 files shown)


yne:
may I ask why `userData` seems more appropriate that just `data` ?
Because it point to a userland address ?

Just askin'

bythos14:
There's not really much difference tbh. I defined it as 'User defined data' so `userData` just seemed like a good name. Honestly could have left it unchanged ¯\_(ツ)_/¯

include/psp2/json.h
```diff
@@ -276,7 +276,7 @@ class Array
 	Array& operator=(const Array& arr);
 
 private:
-	void* impl_data; //!< Pointer to internal implementation data
+	void* impl_data; //!< Pointer to internal implementation data.
```
d3m3vilurr, line 279: userData and errorCode are `camelCase` but `impl_data` is `snake_case`.
could you choose one notation for this lib?

bythos14: Done.

PR #482, opened 2020-10-04
Added SceJpegEncArm enums and edited documentation.

bythos14:
Added missing enums, and adjusted documentation to represent SceJpegEncArm more correctly. Also changed the type of some function args.

### DIFF
--- a/include/psp2/jpegencarm.h
+++ b/include/psp2/jpegencarm.h
@@ -1,101 +1,165 @@
 /**
  * \usergroup{SceJpegEncArm}
  * \internalusage{psp2/jpegencarm.h,SceJpegEncArm_stub,SCE_SYSMODULE_INTERNAL_JPEG_ENC_ARM}
+ *
+ * \note The maximum supported image width and height is 65520 for YCbCr420,
+ *       and 65528 for YCbCr422.\n\n
+ *       The width and height of YCbCr422 images must be a multiple of 8.
+ *       The width and height of YCbCr420 images must be a multiple of 16.
  */
 
 #ifndef _PSP2_JPEGENCARM_H_
 #define _PSP2_JPEGENCARM_H_
 
 #include <psp2/types.h>
-#include <psp2/jpegenc.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#define SCE_JPEGENCARM_MIN_COMP_RATIO     1   //!< Lowest compression ratio, best quality.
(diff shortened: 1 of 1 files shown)


include/psp2/jpegencarm.h
```diff
@@ -1,101 +1,165 @@
...
 int sceJpegArmEncoderEncode(SceJpegArmEncoderContext context, const void *inBuffer);
 
 /**
- * Set encoder compression ratio
+ * Set the encoder compression ratio.
  *
- * @param[in] context - A pointer to an already initialized ::SceJpegArmEncoderContext
- * @param[in] ratio - A value between 0 and 255 (higher = better compression, lower = better speed)
+ * @param[in] context - An already initialized ::SceJpegArmEncoderContext.
+ * @param[in] ratio   - A value between 1 and 256 (higher = better compression, lower = better speed).
```
d3m3vilurr, line 123: 2 to 255?

bythos14: Fixed, was a mistake from early testing.

include/psp2/jpegencarm.h
```diff
@@ -1,101 +1,165 @@
...
 #include <psp2/types.h>
-#include <psp2/jpegenc.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#define SCE_JPEGENCARM_MIN_COMP_RATIO     1   //!< Lowest compression ratio, best quality.
+#define SCE_JPEGENCARM_DEFAULT_COMP_RATIO 64  //!< Default compression ratio.
+#define SCE_JPEGENCARM_MAX_COMP_RATIO     255 //!< Highest compression ratio, lowest quality.
```
d3m3vilurr, line 22: @bythos14 here is 1 to 255. but a document of the `ratio` is 0 to 256.
what is right value?

bythos14: The documentation says **between** 0 and 256, so 0 and 256 are not included in that range of values.

bythos14: So 1 - 255 is the correct range

d3m3vilurr: in normally, `between a and b` is `a <= x <= b`. isn't it?

bythos14: Probably. I guess it would be better to the term inclusively here. I've made the adjustments.

teakhanirons: Documentation from where?

bythos14: The documentation written in the header.

PR #491, opened 2020-10-23
Update SceSblSsUpdateMgr

Princess-of-Sleeping:


d3m3vilurr:
this patch will replace the nid's prefix.
but a header only introduced now.

so, this broken changes would affect limited.

PR #495, opened 2020-10-31
Update SceSysroot nids/functions

Princess-of-Sleeping:


d3m3vilurr:
related #329 
it could be SCE's mistake to give multiple names to one function. and it would reduce the clearity.
I thought our checking process to catch the conflicting hex, but might be lucky, current script only check the conflicting the name.

PR #499, opened 2020-11-01
Add SceSysclib header

Princess-of-Sleeping:


Rinnegatamante:
Can you add the header here too? https://github.com/vitasdk/vita-headers/blob/master/include/vitasdkkern.h

Princess-of-Sleeping:
ok, added.

PR #500, opened 2020-11-05
Update sysclib.h

Princess-of-Sleeping:


d3m3vilurr:
congrats pr number 500.
a patch looks good to me. but it seems to need `-nostdlib` option for compiling.
can you make the simple kernel module sample into vitasdk/samples ?

PR #542, opened 2021-01-12
Add SceCpuSync functions

Princess-of-Sleeping:
from https://github.com/Princess-of-Sleeping/vita-utility/blob/master/CpuCoreSync/src/main.c

### DIFF
--- a/include/psp2kern/kernel/cpu.h
+++ b/include/psp2kern/kernel/cpu.h
@@ -253,6 +253,22 @@ int ksceKernelCpuSpinLockIrqSave(int *addr);
  */
 int ksceKernelCpuSpinLockIrqRestore(int *addr, int flags);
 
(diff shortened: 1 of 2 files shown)


include/psp2kern/kernel/cpu.h
```diff
@@ -253,6 +253,22 @@ int ksceKernelCpuSpinLockIrqSave(int *addr);
  */
 int ksceKernelCpuSpinLockIrqRestore(int *addr, int flags);
 
+#define SCE_CPU_WAIT_CORE_0 1
+#define SCE_CPU_WAIT_CORE_1 2
+#define SCE_CPU_WAIT_CORE_2 3
+#define SCE_CPU_WAIT_CORE_3 0
```
d3m3vilurr, line 259: how about make the `enum`?

PR #577, opened 2021-04-06
Add SCE_KERNEL_ALLOC_MEMBLOCK_ATTR_HAS_ALIGNMENT to user header.

Rinnegatamante:


d3m3vilurr:
we should extract common struct & types.

Rinnegatamante:
I don't know if the other attrs are available in user mode. I know for sure this one is used however (It's required for certain resolutions support with sceAvPlayer)

PR #581, opened 2021-04-19
Add SceSysmoduleOpt struct

Princess-of-Sleeping:


include/psp2/sysmodule.h
```diff
@@ -213,7 +219,7 @@ int sceSysmoduleIsLoadedInternal(SceSysmoduleInternalModuleId id);
  *
```
d3m3vilurr, line 219: need to update a document.

include/psp2/sysmodule.h
```diff
@@ -225,7 +231,7 @@ int sceSysmoduleLoadModuleInternalWithArg(SceSysmoduleInternalModuleId id, SceSi
...
```
d3m3vilurr, line 231: same here

Issue #492, opened 2020-10-30
split the db.yml to muliple files


d3m3vilurr:
an idea from @Princess-of-Sleeping 
current db.yml is too big;

it exists multiple option
- just change the toolchain to support multiple yaml
- just call the toolchain multiple times; but maybe it needs to merge the make files

PR #585, opened 2021-05-13
Split db.yml

Princess-of-Sleeping:
From https://github.com/vitasdk/vita-headers/issues/492

Also experimental.


Princess-of-Sleeping:
One thing to note here is that we have removed the NID prototype check.

https://github.com/vitasdk/vita-headers/pull/585/commits/5d63222fbfcdcfeafaa583a16d3ab1663a0d8113

d3m3vilurr:
replaced #589

PR #592, opened 2021-06-20
Added db for 3.63/3.65/3.67

Electric1447:
Added fl0ws 3.67 db (#300 ) in the new split db format

### DIFF
--- a/db/363/SceKernelModulemgr.yml
+++ b/db/363/SceKernelModulemgr.yml
@@ -1,5 +1,5 @@
 version: 2
-firmware: 3.67
+firmware: 3.63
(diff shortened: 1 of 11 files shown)


Princess-of-Sleeping:
Nice.
But many have changed in 3.63.
For example, SceKernelModulemgr : https://wiki.henkaku.xyz/vita/SceKernelModulemgr#Libraries

So it can be modified from 3.67 to 3.63.

Electric1447:
I'll fix that up later today

Electric1447:
Fixed libraries not being in the correct firmware version

PR #607, opened 2021-07-18
Add libScePiglet NIDs

SonicMastr:


d3m3vilurr:
@frangarcj @Rinnegatamante these nids don't have `sce` prefix unlike other libs. sometimes, it would break thirdparty gl lib. (but that case doesn't link this lib.)
do you want to change `egl..` to `sceEgl...` (eg; `eglBindAPI` to `sceEglBindAPI`) and `gl...` to `sceGl...`(eg; `glActiveTexture` to `sceGlActiveTexture`) or just keep current PR?

frangarcj:
Just keep current. Changing names from standard is not good for this lib.

PR #608, opened 2021-07-21
add/fix known syscalls

Princess-of-Sleeping:
Some documentation or something hasn't been updated correctly and needs to be fixed


Princess-of-Sleeping:
Updating the docs is probably okay, but now I have to update/add the struct

Princess-of-Sleeping:
I built the program with all the user headers included, but it's okay

include/psp2/audioout.h
```diff
@@ -77,16 +77,16 @@ typedef enum SceAudioOutAlcMode {
...
  * @param[in] freq - Sampling frequency (in Hz), one of the followings :
  * 8000, 11025, 12000, 16000, 22050, 24000, 32000, 44100, 48000
- * @param[in] mode - One of ::SceAudioOutMode
+ * @param[in] param - One of ::SceAudioOutMode
  *
  * @return port number, < 0 on error.
  * @note - The volume is initially set to its max value (::SCE_AUDIO_OUT_MAX_VOL)
 */
-int sceAudioOutOpenPort(SceAudioOutPortType type, int len, int freq, SceAudioOutMode mode);
+int sceAudioOutOpenPort(int portType, int len, int freq, int param);
```
d3m3vilurr, line 89: do we need to change the type & need to rename the argument in this file?

include/psp2/avcodec.h
```diff
@@ -0,0 +1,84 @@
...
+SceInt32 _sceVideoencInitLibraryInternal(SceUInt32 codecType, const SceVideodecQueryInitInfoInternal *pInitDecInfo, const SceVideoencQueryInitInfoInternal *pInitEncInfo);
+SceInt32 _sceVideoencInitLibraryWithUnmapMem(SceUInt32 codecType, SceVideoencCtrl *pCtrl, const SceVideodecQueryInitInfo *pInitDecInfo, const SceVideoencQueryInitInfo *pInitEncInfo);
+SceInt32 _sceVideoencInitLibraryWithUnmapMemInternal(SceUInt32 codecType, SceVideoencCtrl *pCtrl, const SceVideodecQueryInitInfoInternal *pInitDecInfo, const SceVideoencQueryInitInfoInternal *pInitEncInfo);
+SceInt32 _sceVideoencQueryMemSize(SceUInt32 codecType, const SceVideodecQueryInitInfo *pInitDecInfo, const SceVideoencQueryInitInfo *pInitEncInfo, SceVideoencMemInfo *pMemInfo);
+SceInt32 _sceVideoencQueryMemSizeInternal(SceUInt32 codecType, const SceVideodecQueryInitInfoInternal *pInitDecInfo, const SceVideoencQueryInitInfoInternal *pInitEncInfo, SceVideoencMemInfo *pMemInfo);
+SceInt32 _sceVideoencTermLibrary(SceUInt32 codecType);
+
+SceInt32 _sceM4vdecCreateDecoder(SceUInt32 codecType, SceM4vdecCtrl *pCtrl, const SceM4vdecQueryDecoderInfo *pDecoderInfo);
```
d3m3vilurr, line 47: can you use `snake_case`? also, we don't need `p` for the pointer.

include/psp2/ctrl.h
```diff
@@ -160,22 +160,22 @@ int sceCtrlSetSamplingModeExt(SceCtrlPadInputMode mode);
...
 
 /**
  * Get the controller state information (polling, positive logic).
  *
- * @param[in] port - use 0.
- * @param[out] *pad_data - see ::SceCtrlData.
- * @param[in] count - Buffers count. Up to 64 buffers can be requested.
+ * @param[in]  port - use 0.
+ * @param[out] pData - see ::SceCtrlData.
+ * @param[in]  nBufs - Buffers count. Up to 64 buffers can be requested.
```
d3m3vilurr, line 174: samely, please restore an argument name if you don't need to change it.

d3m3vilurr (changes requested):
imo, you should rebase & squash commits into merged one commit.
for example, your changes has multiple `Update processmgr.h`

https://company.ona.io/squashing-commits-with-an-interactive-git-rebase/

Princess-of-Sleeping:
All changes to this PR are from the official logs and it is desirable to have the correct function and argument names as much as possible.

Princess-of-Sleeping:
@d3m3vilurr And you often don't like argument names like `pCtrl`, but from a developer's point of view it's clear which name is easier to understand with `ctrl` or `pCtrl`.

Princess-of-Sleeping:
You processed this PR (https://github.com/vitasdk/vita-headers/pull/572) before but didn't say anything about snake_case.

I wondered it.

d3m3vilurr:
first, the best case is one project has one convention.
second, we are in a sad case. our project has mixed. 
third, so now we should choice changing everything after determining the convention rule. or keep the convention on the file (at least)
so, I approved 572 pr that file already had camel cases.

your pr seems more aggressive to change everything before determining the conversation.
so, I request to reduce the changes if we don't really need it yet.

`p` prefix is sometimes helping readability, but I didn't think so because we already use a good editor. and if we used this function with the wrong type variable, a compiler occurs an error.

I mean, a project convention could fix with a conversation between maintainers & contributors. we didn't get it yet.
and maybe it's hard to apply. sometimes it could make breaking changes.

personally, yes I prefer design struct & function naming with snake_case in C. (because I think it's more general in C projects)
but it's related to project's own convention & language.

Princess-of-Sleeping:
Camel cases and snake_case are still controversial, as there are many maintainers and developers (contributors) with their own coding styles.
But I often use both cases.

For example,
```
Camel cases : Fully determined name
snake_case  : Names that have not been determined or do not need to be camelcase (ex: int port) etc.
```

Such things are often found in vita-headers as well.

But as you said, really don't need to rename the SceCtrl function arguments etc. ~~But i think count should be changed to nBufs it's not an important part now~~

The p prefix is very useful for developers using normal editors like me. (i using Notepad)

In any case, I'll undo what I don't need to change, but what do you do with the p prefix in the future?

d3m3vilurr:
ok. I use normal editor too.(vim)
but, I suggest to change to more rich editor like notepad++, vscode and ETC.
they are actually normal editor but they can use plugins :)

actually mixed `camelCase` and `snake_case` is pretty uncommon.
most case uses just one, but somecases use `PascalCase` for the `class` and `structure` name in the `snake_case` world.
another somecases use `camelCase` for the method name.
actually, it's pretty related langauge scene culture.
(for example, I also use camel case variable in java and javascript. and I use pascal case method name in the C#)
just c world is old and nobody knows suggestion of the c lang scene.

anyway. personally, I'm still thinking that we don't need `p` prefix in headers. but it just my opinion. so if you want to use extra prefix or change notation rules, just open the issue into github or have a conversation on the discord channel.

PS. I like this patch things to add function definitions.
so, I just said two things
- change notation cases
- squash splitted commits before merging.

Electric1447:
Amazing!
Maybe we should add functions which might not be present on 3.60 to 3.63/3.65/3.67?

Princess-of-Sleeping:
No, I don't think the function I commented out also exists in 3.61 >=.
Maybe defined in fw 1.69 or less

Princess-of-Sleeping:
This PR will be closed as some commits are messed up and squash can't handle it

PR #636, opened 2021-09-05
Add libime.h / Fix ime_dialog.h

SonicMastr:


frangarcj:
What is the reason of this change?

SonicMastr:
`libime` functions currently have no definitions, nor their own header. `libime` is a the "mother" library for `ime_dialog` and some of the types present in `ime_dialog.h` actually belong to `libime`. 

This adds the definitions for the missing `libime` functions while also reorganizing definitions to be in their correct headers (So there aren't any duplicates in places where they're not supposed to be). It also adds `libime` error codes and some extra `SceImeTypes` (`SCE_IME_TYPE_URL` and `SCE_IME_TYPE_MAIL`) that were previously not added.

It's really just an addition + cleanup to better represent how `ime_dialog` is dependent on `libime`, but `libime` is still its own independent library.

include/psp2/libime.h
```diff
@@ -0,0 +1,185 @@
+/**
+ * \usergroup{SceIme}
+ * \usage{psp2/libime.h,SceIme_stub,SCE_SYSMODULE_IME}
```
d3m3vilurr, line 3: how about just `ime.h`?

d3m3vilurr: nvm i forgot we already have `libdbg.h` and `libssl.h`

but i just thought `ime_dialog.h` & `libime.h` are pretty related in here.
(similarly, `common_dialog.h`  and `ime_dialog.h` are also related)
so i pass the decision role to other maintainers and @SonicMastr

SonicMastr: I also tried checking to see if any other header had the `lib` prefix before settling with `libime.h`. I stuck with it because I thought it would be a bit more accurate. It wouldn't be an issue to change it to `ime.h` if you want, though.

frangarcj: it's not bad

SonicMastr:
Anything else I need to do in order for this to me merged? It'll be necessary for the next SDL update

PR #639, opened 2021-09-11
Fix SceCorelockCore enum

Princess-of-Sleeping:


include/psp2kern/kernel/cpu.h
```diff
@@ -254,16 +254,16 @@ int ksceKernelCpuSpinLockIrqSave(int *addr);
 int ksceKernelCpuSpinLockIrqRestore(int *addr, int flags);
 
 typedef enum SceCorelockCore {
-	SCE_CORELOCK_CORE0 = 1,
-	SCE_CORELOCK_CORE1 = 2,
-	SCE_CORELOCK_CORE2 = 3,
-	SCE_CORELOCK_CORE3 = 0
+	SCE_CORELOCK_CORE0 = 0x00000000,
```
d3m3vilurr, line 257: could you value change to 0~3 instead hex?

d3m3vilurr (changes requested):

include/psp2kern/kernel/cpu.h
Princess-of-Sleeping: done

PR #647, opened 2021-09-30
Complete libMonoBridge header.

KuromeSan:
This should make messing around with PSM way easier..

### DIFF
--- a/include/psp2/pss.h
+++ b/include/psp2/pss.h
@@ -8,6 +8,7 @@
 #define _PSP2_PSS_H_
 
 #include <psp2/types.h>
+#include <pthread.h>
 
 #ifdef __cplusplus
 extern "C" {
--- a/include/psp2/pss.h
+++ b/include/psp2/pss.h
@@ -15,12 +15,11 @@ extern "C" {
 #endif
 
 #define PSS_OK 0
-#define PSS_ERROR_THREAD_SUSPENDED -2147319763
+#define PSS_ERROR_THREAD_SUSPENDED 0x8002802D
 
-/* IO Defines */
-#define PSS_ERROR_ERRNO_EEXIST -2147418095
-#define PSS_ERROR_ERRNO_ENOENT -2147418110
-#define PSS_ERROR_ERRNO_EACCES -2147418099
+#define PSS_ERROR_ERRNO_EEXIST 0x80010011
+#define PSS_ERROR_ERRNO_ENOENT 0x80010002
+#define PSS_ERROR_ERRNO_EACCES 0x8001000D
 
 #define PSS_STM_RWU (00600)
 #define PSS_STM_FDIR (0x1 << 12)
@@ -36,7 +35,6 @@ extern "C" {
 #define PSS_CST_AT 0x0010
 #define PSS_CST_MT 0x0020
 
-/* Net defines */
 #define PSS_NET_IPPROTO_IP            0
 #define PSS_NET_IPPROTO_ICMP          1
 #define PSS_NET_IPPROTO_IGMP          2
@@ -274,40 +272,40 @@ typedef struct pss_ucontext {
 	PSSFloatRegisters fregs;
 } pss_ucontext;
 
-typedef struct _PssCryptoContext {
+typedef struct PssCryptoContext {
 	int handle;
 	int valid;
 	int size;
 	int type;
 } PssCryptoContext;
 
-typedef struct _PssTimeval {
-	long tv_sec;
-	long tv_usec;
+typedef struct PssTimeval {
+	int64_t tv_sec;
+	int64_t tv_usec;
 } PssTimeval;
 
-typedef struct _PssTimespec {
+typedef struct PssTimespec {
 	int tv_sec;
 	int tv_nsec;
 } PssTimespec;
 
-typedef struct _PssTimezone {
+typedef struct PssTimezone {
 	int tz_minuteswest;
 	int tz_dsttime;
 } PssTimezone;
 
-typedef struct _PssDateTime {
-    unsigned short year;
-    unsigned short month;
-    unsigned short day;
-    unsigned short hour;
-    unsigned short minute;
-    unsigned short second;
+typedef struct PssDateTime {
+    uint16_t year;
+    uint16_t month;
+    uint16_t day;
+    uint16_t hour;
+    uint16_t minute;
+    uint16_t second;
     unsigned int   microsecond;
 } PssDateTime;
 
 
-typedef struct _PssIoStat {
+typedef struct PssIoStat {
 	int st_mode;
 	unsigned int st_attr;
 	int64_t st_size;
@@ -317,31 +315,31 @@ typedef struct _PssIoStat {
 	unsigned int st_private[6];
 } PssIoStat;
 
-typedef struct _PssNetHostent {
-    char  *h_name;
+typedef struct PssNetHostent {
+    char *h_name;
     char **h_aliases;
-    int    h_addrtype;
-    int    h_length;
+    int h_addrtype;
+    int h_length;
     char **h_addr_list;
 } PssNetHostent;
 
-typedef struct _PssNetLinger {
+typedef struct PssNetLinger {
     int l_onoff;
     int l_linger;
 } PssNetLinger;
 
-typedef struct _PssNetSockaddr {
+typedef struct PssNetSockaddr {
     unsigned char sa_len;
     unsigned char sa_family;
     signed char sa_data[14];
 } PssNetSockaddr;
 
-typedef struct _PssNetIovec {
+typedef struct PssNetIovec {
     void *iov_base;
     int iov_len;
 } PssNetIovec;
 
-typedef struct _PssNetMsghdr {
+typedef struct PssNetMsghdr {
     void *msg_name;
     uint32_t msg_namelen;
     PssNetIovec *msg_iov;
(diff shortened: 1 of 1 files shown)


KuromeSan:
fix your cmake >_<

include/psp2/pss.h
```diff
@@ -8,29 +8,566 @@
...
-
-int pss_crypto_open(ScePssCryptoHandle *handle, char *path);
-char *pss_crypto_read(ScePssCryptoHandle *handle, int *mode);
-int pss_crypto_close(ScePssCryptoHandle *handle);
-void *pss_code_mem_alloc(SceSize *);
-void pss_code_mem_flush_icache(const void *, SceSize);
-void pss_code_mem_lock(void);
-void pss_code_mem_unlock(void);
+#define PSS_OK 0
+#define PSS_ERROR_THREAD_SUSPENDED -2147319763
```
d3m3vilurr, line 18: `0x8002802D` or use `SCE_KERNEL_ERROR_THREAD_SUSPENDED` in `kernel/error.h`

include/psp2/pss.h
```diff
@@ -8,29 +8,566 @@
...
+typedef struct _PssCryptoContext {
+	int handle;
+	int valid;
+	int size;
+	int type;
+} PssCryptoContext;
+
+typedef struct _PssTimeval {
+	long tv_sec;
+	long tv_usec;
```
d3m3vilurr, line 286: can you use `int64_t / uint16_t` instead of `long/short`?

include/psp2/pss.h
```diff
@@ -8,29 +8,566 @@
...
+#define PSS_ERROR_ERRNO_EACCES -2147418099
+
+#define PSS_STM_RWU (00600)
+#define PSS_STM_FDIR (0x1 << 12)
+#define PSS_STM_FREG (0x2 << 12)
+
+#define PSS_FREAD (0x0001)
+#define PSS_FWRITE (0x0002)
+#define PSS_FCREAT (0x0200)
+#define PSS_TRUNC (0x0400)
```
d3m3vilurr, line 32: maybe we should extract enums in `net.h`, `fcntl.h` and others to spererate files

at least, could you make the enum instead defines?

include/psp2/pss.h
```diff
@@ -13,24 +13,560 @@
...
+	PssDateTime st_ctime;
+	PssDateTime st_atime;
+	PssDateTime st_mtime;
+	unsigned int st_private[6];
+} PssIoStat;
+
+typedef struct _PssNetHostent {
+    char  *h_name;
+    char **h_aliases;
+    int    h_addrtype;
```
d3m3vilurr, line 323: remove extra spaces between type and name

include/psp2/pss.h
```diff
@@ -13,24 +13,560 @@
...
+	int st_mode;
+	unsigned int st_attr;
+	int64_t st_size;
+	PssDateTime st_ctime;
+	PssDateTime st_atime;
+	PssDateTime st_mtime;
+	unsigned int st_private[6];
+} PssIoStat;
+
+typedef struct _PssNetHostent {
```
d3m3vilurr, line 320: most type & enum are same as toolchain's or another headers.
so imo good option is
```
#include <netdb.h>
typedef struct hostent PssNetHostent;
```
but i'm not sure we can do possible cross ref newlib's header in here.

include/psp2/pss.h
```diff
@@ -8,29 +8,566 @@
...
+		uint32_t word[4];
+	} reg[16];
+} PSSFloatRegisters;
+
+typedef struct pss_ucontext {
+	PSSIntegerRegisters iregs;
+	PSSFloatRegisters fregs;
+} pss_ucontext;
+
+typedef struct _PssCryptoContext {
```
d3m3vilurr, line 277: why it has renamed and why it has `_` prefix?

include/psp2/pss.h
```diff
@@ -8,29 +8,566 @@
...
+/* Threading */
+typedef void (*pss_exception_callback)(pthread_t tid, pss_ucontext *ctx);
+int32_t pss_suspend_thread (int tid, uint32_t *timeout);
+int32_t pss_resume_thread (int tid);
+int32_t pss_set_thread_context (int tid, PSSIntegerRegisters *iregs, PSSFloatRegisters *fregs);
+int32_t pss_get_thread_context (int tid, PSSIntegerRegisters *iregs, PSSFloatRegisters *fregs);
+void pss_disable_ftz ();
+int pss_supports_fast_tls ();
+void pss_threads_initialize (pss_exception_callback cb);
+void *pss_get_stack_bottom ();
```
d3m3vilurr, line 411: functions has extra spaces between name and `(`. remove it.

include/psp2/pss.h
KuromeSan: Offical name, saw it like SceDebugPrintf("Invalid PssCryptoContext")

include/psp2/pss.h
```diff
@@ -8,29 +8,566 @@
...
+#define PSS_NET_EPOLL_CTL_DEL   3
+
+typedef struct PssIRegisters {
+	unsigned int size;
+	uint32_t cpsr;
+	uint32_t reg[16];
+	uint32_t tpidrurw;
+	uint32_t teehbr;
+	uint32_t stackBottom;
+	uint32_t stackTop;
```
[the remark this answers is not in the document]
KuromeSan: this is the mono name though.. https://github.com/OpenPSS/pss-mono/blob/master/include/bridge.h#L268 look, this entire header has just been sitting in there GPL source release for like 10 years, :D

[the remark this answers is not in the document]
KuromeSan: this is the name from the mono VM, its kind of weird that they just dont use snake case here, but thats how it works.

include/psp2/pss.h
KuromeSan: PssCryptoContext is the official name ..

include/psp2/pss.h
KuromeSan: hm, would have to look through the entire SDK and find where any match?

include/psp2/pss.h
```diff
@@ -8,29 +8,566 @@
...
+	    (int)PSS_NET_ID_SOCKET_MIN, (int)PSS_NET_ID_SOCKET_MAX)
+#define FD_CLR(n, p) \
+	PSS_NET_ID_CLR(n, p, \
+	    (int)PSS_NET_ID_SOCKET_MIN, (int)PSS_NET_ID_SOCKET_MAX)
+#define FD_ISSET(n, p) \
+	PSS_NET_ID_ISSET(n, p, \
+	    (int)PSS_NET_ID_SOCKET_MIN, (int)PSS_NET_ID_SOCKET_MAX)
+#define FD_ZERO(p) \
+	PSS_NET_ID_ZERO(p, PSS_NET_FD_SETSIZE)
+#define FD_SETSIZE PSS_NET_FD_SETSIZE
```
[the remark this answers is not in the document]
KuromeSan: from pss mono code

include/psp2/pss.h
KuromeSan: im not sure what you mean ? move these defines to net.h?

d3m3vilurr: i mean we should split enums to a sperate file from `net.h` then share this values.

include/psp2/pss.h
d3m3vilurr: i'm not sure do we need to define twice?5

include/psp2/pss.h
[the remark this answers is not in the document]
d3m3vilurr: it's okay. could you share reference parts information of the mono base codeset into commit message or comment?
some functions i've found of it like `https://github.com/mono/mono/blob/main/support/map.h`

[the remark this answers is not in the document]
KuromeSan: Specifically in sony Mono PSM Mono fork
https://github.com/OpenPSS/pss-mono/blob/mono-2-10/include/bridge.h
In bridge.h lol ~~which is where most of this comes from~~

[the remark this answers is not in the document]
d3m3vilurr:
```
Copyright (C) 2012, SCEA LLC
```

first, `openpss` uses dual license with `GPL` & `LGPL`. it couldn't be possible to use into here without maintainer/original author's call. also it could be bad point of it. because mono is MIT.
second, imo it has gray area of sce part; at least it should be `LGPL v2` (see the `LICENSE_SCE`. but I need actual reference. (~~really did sony release it with lgpl?~~ https://doc.dl.playstation.net/doc/psvita-oss/mono.html )

if sony was released(or leaked) the original source code with lgpl, the `bridge.h` header have to apply `lgpl`
so technically it's lgpl(and it's reason why openpss is lgpl), but a problem of first case is still remain.
and imo, sony doesn't agree to change the licence of the MIT for vitasdk project :p

so my opinion is split the project for the mono header under lgpl instead adding into here.
(or we have to update the license part then adding header with license description)

KuromeSan:
to properly implement libmonobridge some functions take MonoArray* and MonoString* as arguments; (see for example AccountInformation  implementation in the SnowPME Emulator, https://github.com/OpenPSS/SnowPME/blob/53335f8a7bc8fee471fc56c13f6f2c5868dec07b/LibPSM/Sce/PlayStation/Core/Services/AccountInformation.cpp#L10 
which would require including some of mono's headers. 

also technically LibMonoBridge isn't a system library and is part of a 3rd party application (PSM Runtime & PSM Dev) and its inclusion is probably a remnant from the Rejuvinate days; but its useful to have this around for i.e reverse engineering or plugins/homebrew that runs in context of the PSM Runtime (which there are a few .. ) 

this PR has been sitting dormant for a few years but i don't think its really possible to add this to the SDK for these reasons.

PR #653, opened 2021-10-12
remove unk2

KuromeSan:
i dont think this parameter exists.. 
see: https://github.com/KuromeSan/vshPatch/blob/0e4823cc3bb860b66795fc02ac66e248e07fcf28/vshPatch.c#L15
given that this actually works, when it probably would not if this unk2 was really here,


d3m3vilurr:
I wouldn't merge of it.
1. it seems field alignment issue of the compiler. imo, a best option is keep the structure & add `__attribute__((packed))` into structure definition
2. so now it doesn't need `union` & `reserved`  field

PR #660, opened 2021-10-16
add scePsmDrmGetRifKey and psm license struct

EliCrystal2001:
Maybe should goto a separate psmdrm.h, not sure.

### DIFF
--- a/include/psp2/npdrm.h
+++ b/include/psp2/npdrm.h
@@ -39,20 +39,20 @@ typedef struct SceNpDrmLicense { // size is 0x200
 
 typedef struct ScePsmDrmLicense {
   char magic[0x8];             
-  uint32_t unk1;               
-  uint32_t unk2;               
-  uint64_t aid;                
-  uint32_t unk3;               
-  uint32_t unk4;               
-  uint64_t start_time;         
-  uint64_t expiration_time;    
-  uint8_t act_sha256digest[0x20];    
+  SceUInt32 unk1;               
+  SceUInt32 unk2;               
+  SceUInt64 account_id;                
+  SceUInt32 unk3;               
+  SceUInt32 unk4;               
+  SceUInt64 start_time;         
+  SceUInt64 expiration_time;    
+  SceUInt8 activation_checksum[0x20];    
   char content_id[0x30];       
-  uint8_t unk5[0x80];          
-  uint8_t unk6[0x20];
-  uint8_t key[0x10];
-  uint8_t signature[0x1D0];
-  uint8_t signature_rsa[0x100]; 
+  SceUInt8 unk5[0x80];          
+  SceUInt8 unk6[0x20];
+  SceUInt8 key[0x10];
+  SceUInt8 signature[0x1D0];
+  SceUInt8 rsa_signature[0x100]; 
 } ScePsmDrmLicense;
   
 /**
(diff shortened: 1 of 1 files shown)


Princess-of-Sleeping:
sha256digest is maybe RSA sign?

EliCrystal2001:
i dont know, i got this struct from NoPsmDrm.

d3m3vilurr:
@frangarcj can you allow to change the license of the nopsmdrm's structure?

KuromeSan:
this struct is wrong, key is only 0x10 bytes, see: https://github.com/KuromeSan/psse-decrypt/blob/master/decrypt.py#L28
it should be like unk[0x20] key[0x10] signature[0x1C0]

also PSM usually uses MD5 not SHA256, so Princess of Sleeping is probably correct about that being wrong.

frangarcj:
> @frangarcj can you allow to change the license of the nopsmdrm's structure?

No problem

include/psp2/npdrm.h
```diff
@@ -37,6 +37,24 @@ typedef struct SceNpDrmLicense { // size is 0x200
...
+typedef struct ScePsmDrmLicense {
+  char magic[0x8];             
+  uint32_t unk1;               
+  uint32_t unk2;               
+  uint64_t aid;                
+  uint32_t unk3;               
+  uint32_t unk4;               
+  uint64_t start_time;         
+  uint64_t expiration_time;    
+  uint8_t act_sha256digest[0x20];
```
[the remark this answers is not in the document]
KuromeSan: offical act sha256sum: 76b2cb81a31f07d18afc6ae903742e717510eb642d9fd4e744ce69012113b951 
offical rif 0x30 : E5E2511CF85B3264DB8B1FE30E8075657524BFF26E3F0452C60973566B613AF0

doenst seem like a sha256sum, possibly hmac?

SonicMastr:
> Maybe should goto a separate psmdrm.h, not sure.

I actually think that due to the fact that it's part of a separate ScePsmDrm library, it probably should. I guess that's really up to @d3m3vilurr though

EliCrystal2001:
> > Maybe should goto a separate psmdrm.h, not sure.
> 
> I actually think that due to the fact that it's part of a separate ScePsmDrm library, it probably should. I guess that's really up to @d3m3vilurr though

however its still in NpDrm.skprx.

SonicMastr:
Yes. But having the same header for multiple libraries generally isn't good practice. A module can either be a single library or a collection of libraries and just having them separated would make it more intuitive to use. Not all headers in VitaSDK are named specifically after the module but as the sub-libraries within them to make it easier to differentiate the underlying libraries and functions. It's more of a clarity thing.

include/psp2/npdrm.h
```diff
@@ -72,6 +90,19 @@ int _sceNpDrmGetFixedRifName(char *rif_name, uint64_t aid);
...
+/**
+ * Get PSM rif key
+ *
+ * @param[in] license_buf - RIF buffer (1024 bytes)
+ *
+ * @param[out] keydata - Decrypted key data
+ *
+ * @param[in] flags - Unknown
+ * 
+ * @return >0 on success, <0 on error
```
d3m3vilurr, line 102: is it right? most time 0 is normal.

EliCrystal2001: your right it is =0.

PR #673, opened 2021-11-20
Update paf

Princess-of-Sleeping:
Add some memory stuff and rename the stdc function to the name given by the bruteforce attack.
But now Removed unnecessary sce_paf_private_malloc2 and sce_paf_private_free2.

sha1("memcpyPAF") = `e3a68358 50eec820606f2aad1c617c18fa21598f`.

The remaining `sce_paf_private_*` functions are those not found in the bruteforce attack.

### DIFF
--- a/.travis.d/definition_check.py
+++ b/.travis.d/definition_check.py
@@ -70,6 +70,8 @@ def read_nids():
             for line_no, line in enumerate(readlines(d)):
                 line = line.strip()
                 k, v = line.split(':')[:3]
+                if k[0] == "#": # It a comment
+                    continue
                 if not v.strip():
                     SECTION = k
                     continue
--- a/db/360/ScePaf.yml
+++ b/db/360/ScePaf.yml
@@ -8,7 +8,13 @@ modules:
         kernel: false
         nid: 0x3D643CE8
         functions:
+          scePafCreateHeap: 0xA2B0BF4D
+          scePafDeleteHeap: 0x25238A3A
+          scePafFreeWithContext: 0xCE97B579
           scePafGetCurrentClockLocalTime: 0x96345146
+          scePafMallocAlignWithContext: 0x04716F15
+          scePafMallocWithContext: 0x89A1CE78
+          scePafReallocWithContext: 0x74DB7F5D
           scePafSha1Init: 0xB38C53C0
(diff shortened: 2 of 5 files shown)


.travis.d/definition_check.py
```diff
@@ -80,7 +80,7 @@ def read_nids():
                         nids = user_nids
                 if SECTION != 'functions':
                     continue
-                if nids.get(k):
+                if nids != user_nids and nids.get(k):
```
d3m3vilurr, line 83: is it safe to check ther user_nids?

Princess-of-Sleeping: I think no.
Or list only really common functions and allow skipping only if the target function is duplicated on list.
In here, just memcpy/memset to list.

Princess-of-Sleeping: Added a patch to checks duplicate names for each module.

.travis.d/definition_check.py
```diff
@@ -80,9 +82,10 @@ def read_nids():
                         nids = user_nids
                 if SECTION != 'functions':
                     continue
-                if nids.get(k):
+                if current_nids.get(k):
```
d3m3vilurr, line 85: imo both are needed.
```
if nids.get(k):
  errors.append(...)
elif current_nids.get(k):
  errors.append(...)
```
but, actually i want to remove ambiguous conflicting case between libs.
after the merging this PR, we can't load ScePaf and SceLibc at sametime
also compiler requires nostdlib.

so my opinion is just change the function name of the ScePaf.
for example; `memcpy/memset` to `sce_paf_memcpy/memset`(or `sce_paf_libc_memcpy/memset`) or just don't add them to above conflicting with `0x2C5B6F9C/0xE148AF94`
imo we only need one of them. isn't it?

Princess-of-Sleeping: Yes. I have never seen such a case at SCE.
But, if project requires ScePaf and want to include external library sources, all memcpy/memset must be rewritten to sce_paf_memcpy/sce_paf_memset.

But now it doesn't matter. So memcpy/memset will be commented out for the future.

Princess-of-Sleeping:
Ready to merge.

PR #680, opened 2022-01-11
Add SCE_KERNEL_THREAD_ID_SELF

CreepNT:
Also removed comments of confusion related to missing define. May have missed some.


CreepNT:
I fricked up the Git history, fixing and reopening.

PR #683, opened 2022-01-22
Fix layout of SceShaccCgCompileOptions

bythos14:


d3m3vilurr:
@bythos14 is this safe/right offset?
some applications uses old structure format.
for example; [VitaShaRK](https://github.com/Rinnegatamante/vitaShaRK/blob/30e2361/source/vitashark.c#L95-L101)

bythos14:
It's the right offset. I've tested and it behaves as it should.

isage:
I can confirm that it's right offset. shacccg is a cut-down version of psp2cgc, and uses same option struct. REing psp2cgc allows to correctly map command-line options to struct fields.

PR #701, opened 2022-04-21
fix: psp2 dependency in psp2kern

ZetaNumbers:


include/psp2kern/avcodec/jpegenc.h
```diff
@@ -6,7 +6,7 @@
 #ifndef _PSP2_AVCODEC_JPEGENC_H_
 #define _PSP2_AVCODEC_JPEGENC_H_
```
d3m3vilurr, line 7: additionally, can you change this definition check to

```suggestion
#ifndef _PSP2_KERNEL_AVCODEC_JPEGENC_H_
#define _PSP2_KERNEL_AVCODEC_JPEGENC_H_
```

ZetaNumbers: Oh, i think KERNEL prefix collides with psp2/kernel headers definition check prefix. Hm, should i make another PR or fix it here?

ZetaNumbers: There are around ten of other cases like this too.

d3m3vilurr: sure. you can make another pr :)

PR #702, opened 2022-04-24
Consistent header guards

ZetaNumbers:
Came up in #701. Fixes header guard collisions. Corrects and adds `#endif /* <guard name> */` comments.

This PR renames header guards, according to these rules:
1. Directory names must be lowercase and contain only one word;
2. Header file names can contain multiple words separated with underscores, but otherwise are the same as directory names;
3. Header guard names are to be written in uppercase and mimic the path from include directory of the repo.

It allows current naming for header files, but defines consistent naming for the guards. However it would allow guard collisions for files like `module_sub.h` and `module/sub.h`, but that would be not very plausible, i believe. However it isn't required to enforce these rules in the future, i have defined them just for this PR.

### DIFF
--- a/include/psp2/apputil.h
+++ b/include/psp2/apputil.h
@@ -4,8 +4,8 @@
  */
 
 
-#ifndef _PSP2_APPUTL_H_
-#define _PSP2_APPUTL_H_
+#ifndef _PSP2_APPUTIL_H_
+#define _PSP2_APPUTIL_H_
 
 #include <psp2/types.h>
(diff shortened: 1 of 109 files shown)


ZetaNumbers:
All hail nushell!

It helped a lot to replace all of it automatically. Do not worry, i have checked them after that.

ZetaNumbers (commented):
All hail nushell!

It helped a lot to replace all of it automatically. Do not worry, i have checked them after that.

PR #725, opened 2022-07-16
fix: `SceKernelTimeval` fields

ZetaNumbers:
Based upon boost's `xtime` definition and implementation of `_Thrd_sleep`.


Princess-of-Sleeping:
Even on SceRtc, it was certainly `SceUInt32[2]`.

PR #727, opened 2022-07-26
`RWLock` definitions

ZetaNumbers:
A little bit of borrowing from vita3k, reverse engineering, a little of extrapolation and testing.

Does not add `sceKernelCancelRWLock` because why.

### DIFF
--- a/include/psp2/kernel/threadmgr.h
+++ b/include/psp2/kernel/threadmgr.h
@@ -1246,7 +1246,16 @@ typedef struct SceKernelRWLockInfo {
 	SceUID          rwlockId;
 	/** NULL-terminated name of the rwlock */
 	char            name[32];
-	int             unk[5];
+	/** Attributes */
+  SceUInt32       attr;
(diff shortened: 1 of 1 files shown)


include/psp2/kernel/threadmgr.h
```diff
@@ -1228,6 +1228,146 @@ void *sceKernelGetThreadTLSAddr(SceUID thid, int key);
...
+ * @see sceKernelGetRWLockInfo
+ */
+typedef struct SceKernelRWLockInfo {
+	/** Size of the ::SceKernelRWLockInfo structure */
+	SceSize         size;
+	/** The UID of the rwlock */
+	SceUID          rwlockId;
+	/** NULL-terminated name of the rwlock */
+	char            name[32];
+	int             unk[5];
```
isage, line 1249: Don't use non-fixed size types.
Also, that structure described on wiki: https://wiki.henkaku.xyz/vita/SceKernelThreadMgr

PR #733, opened 2022-09-22
Tweak defs.h

Princess-of-Sleeping:
Add ifndef to make it more portable.
Also move defs.h to psp2common from psp2 since may be used by kernel.


Princess-of-Sleeping:
Do not merge yet

PR #736, opened 2022-10-30
Create SceKernelBootimage.yml

Princess-of-Sleeping:


d3m3vilurr:
we should update ci chain to use mirrors.
just merge it now.

PR #739, opened 2022-11-29
Fixed modulemgr stuff

Princess-of-Sleeping:


include/psp2kern/kernel/processmgr.h
```diff
@@ -98,7 +80,7 @@ SceUID ksceKernelGetProcessMainThread(SceUID pid);
  *
  * @return      The process module cb pointer
  */
-SceKernelProcessModuleInfo *ksceKernelGetProcessModuleInfo(SceUID pid);
+ScePVoid *ksceKernelGetProcessModuleInfo(SceUID pid);
```
d3m3vilurr, line 83: is it right to return `void **`?

Princess-of-Sleeping: oops, it mistake.

PR #740, opened 2022-12-01
Usb

isage:
* Add missing nids
* Properly name functions
* Add SceUsbserv headers
* Fix SceUsbd headers
current ones are partly copypasted from ps4 sdk, which is wrong, since ps4 uses totally different api (libusb), vita uses same api as ps2/ps3
those come from RE and usb specs

### DIFF
--- a/include/psp2/usbd.h
+++ b/include/psp2/usbd.h
@@ -7,89 +7,505 @@
 #ifndef _PSP2_USBD_H_
 #define _PSP2_USBD_H_
 
-#include <psp2/kernel/threadmgr.h>
-
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 typedef enum SceUsbdErrorCode {
-	SCE_USBD_ERROR_IO            = 0x80240001,
-	SCE_USBD_ERROR_INVALID_ARG   = 0x80240002,
-	SCE_USBD_ERROR_ACCESS        = 0x80240003,
-	SCE_USBD_ERROR_NO_DEVICE     = 0x80240004,
-	SCE_USBD_ERROR_NOT_FOUND     = 0x80240005,
-	SCE_USBD_ERROR_BUSY          = 0x80240006,
-	SCE_USBD_ERROR_TIMEOUT       = 0x80240007,
-	SCE_USBD_ERROR_OVERFLOW      = 0x80240008,
-	SCE_USBD_ERROR_PIPE          = 0x80240009,
-	SCE_USBD_ERROR_INTERRUPTED   = 0x8024000A,
-	SCE_USBD_ERROR_NO_MEM        = 0x8024000B,
-	SCE_USBD_ERROR_NOT_SUPPORTED = 0x8024000C,
-	SCE_USBD_ERROR_FATAL         = 0x802400FF
+    SCE_USBD_ERROR_NOT_INITIALIZED     = 0x80240001,
+    SCE_USBD_ERROR_ALREADY_INITIALIZED = 0x80240002,
+    SCE_USBD_ERROR_INVALID_PARAM       = 0x80240003,
+
+    SCE_USBD_ERROR_PIPE_NOT_FOUND      = 0x80240004,
+
+    SCE_USBD_ERROR_NO_MEMORY           = 0x80240005,
+    SCE_USBD_ERROR_DEVICE_NOT_FOUND    = 0x80240006,
+
+    SCE_USBD_ERROR_80240007            = 0x80240007, //
+    SCE_USBD_ERROR_80240009            = 0x80240009, //
+    SCE_USBD_ERROR_8024000A            = 0x8024000A, //
+
+    SCE_USBD_ERROR_FATAL               = 0x802400FF
 } SceUsbdErrorCode;
 
 typedef struct SceUsbdDeviceInfo {
-	unsigned int unk0;
-	unsigned int device_id;
-	unsigned int unk2;
-} SceUsbdDeviceInfo; /* size = 0xC */
-
-typedef struct SceUsbdDeviceAddress {
-	unsigned int unk0;
-	unsigned short unk1;
-} SceUsbdDeviceAddress; /* size = 0x6 */
+    unsigned int port;
+    unsigned int device_num;
+    unsigned int unk3; // handled? 0, 1, 2
+} SceUsbdDeviceInfo;
 
 typedef struct SceUsbdTransferData {
-	unsigned int pipe;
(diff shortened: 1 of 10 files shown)


include/psp2/usbserv.h
```diff
@@ -0,0 +1,41 @@
...
+#define _PSP2_USBSERV_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef enum SceUsbservErrorCode {
+    SCE_USBSERV_ERROR_UNAVAILABLE    = 0x80248002,
+    SCE_USBSERV_ERROR_INVALID_PARAM  = 0x80248003,
+    SCE_USBSERV_ERROR_NOT_SUPPPORTED = 0x80248004,
```
d3m3vilurr, line 16: typo

include/psp2/usbd.h
```diff
@@ -7,89 +7,505 @@
...
+    uint8_t wDescriptorLength1;
+} SceUsbdHidSubDescriptorInfo;
+
+typedef struct SceUsbdHidDescriptor {
+    uint8_t bLength;
+    uint8_t bDescriptorType;
+    uint8_t bcdHID0;
+    uint8_t bcdHID1;
+    uint8_t bCountryCode;
+    uint8_t bNumDescriptors;  /* SubDescriptor count */
```
d3m3vilurr, line 247: can you change these field variable names to `snake_case`?

include/psp2/usbd.h
```diff
@@ -7,89 +7,505 @@
...
+    unsigned int unk2; // != 8, set to 1, 2, 4? // transfer flags? type?
+    unsigned int unk3; // ??
+    unsigned int unk4; // ??
+    unsigned int unk5; // ??
+    unsigned int transfer_id; // ??
 } SceUsbdReceiveEvent; /* size = 0x1C */
 
+typedef struct SceUsbdDeviceAddress {
+        unsigned int unk0;
+        unsigned short unk1;
```
d3m3vilurr, line 59: it should be 4 space indent like other structurs.

include/psp2/usbd.h
```diff
@@ -7,89 +7,505 @@
...
+#define USB_DESCRIPTOR_DEVICE                   0x01    // bDescriptorType for a Device Descriptor.
+#define USB_DESCRIPTOR_CONFIGURATION            0x02    // bDescriptorType for a Configuration Descriptor.
+#define USB_DESCRIPTOR_STRING                   0x03    // bDescriptorType for a String Descriptor.
+#define USB_DESCRIPTOR_INTERFACE                0x04    // bDescriptorType for an Interface Descriptor.
+#define USB_DESCRIPTOR_ENDPOINT                 0x05    // bDescriptorType for an Endpoint Descriptor.
+
+#define USB_DESCRIPTOR_DEVICE_QUALIFIER         0x06    // bDescriptorType for a Device Qualifier.
+#define USB_DESCRIPTOR_OTHER_SPEED              0x07    // bDescriptorType for a Other Speed Configuration.
+#define USB_DESCRIPTOR_INTERFACE_POWER          0x08    // bDescriptorType for Interface Power.
+#define USB_DESCRIPTOR_OTG                      0x09    // bDescriptorType for an OTG Descriptor.
```
d3m3vilurr, line 71: can we define enum for these?

include/psp2/usbd.h
```diff
@@ -7,89 +7,505 @@
...
+#define SCE_USBD_CLASS_COMMUNICATIONS                0x02
+#define SCE_USBD_CLASS_HID                           0x03
+#define SCE_USBD_CLASS_MONITOR                       0x04
+#define SCE_USBD_CLASS_PHYSICAL                      0x05
+#define SCE_USBD_CLASS_POWER                         0x06
+#define SCE_USBD_CLASS_PRINTER                       0x07
+#define SCE_USBD_CLASS_STORAGE                       0x08
+#define SCE_USBD_CLASS_HUB                           0x09
+#define SCE_USBD_CLASS_DATA                          0x0a
+#define SCE_USBD_CLASS_VENDOR_SPECIFIC               0xff
```
d3m3vilurr, line 141: can we define enum for these?

include/psp2/usbd.h
```diff
@@ -7,89 +7,505 @@
...
+    SCE_USBD_ERROR_PIPE_NOT_FOUND      = 0x80240004,
+
+    SCE_USBD_ERROR_NO_MEMORY           = 0x80240005,
+    SCE_USBD_ERROR_DEVICE_NOT_FOUND    = 0x80240006,
+
+    SCE_USBD_ERROR_80240007            = 0x80240007, //
+    SCE_USBD_ERROR_80240009            = 0x80240009, //
+    SCE_USBD_ERROR_8024000A            = 0x8024000A, //
+
+    SCE_USBD_ERROR_FATAL               = 0x802400FF
```
d3m3vilurr, line 30: are old names wrong?
and also maybe we need to define `deprecated` marking

isage: Old names are copypasted from ps4 and are totally wrong

d3m3vilurr:
and if you can do, please rebase the commits to remove `typo` commits..

include/psp2/usbd.h
isage: No. That's how they names on every other system. Renaming them just makes porting harder.

include/psp2/usbd.h
isage: No

include/psp2/usbd.h
isage: No

d3m3vilurr:
ah got it. sorry, i thought wrongly. you mean, current error codes could be copied from ps4sdk or others. and it would be wrong.
now, you check/dig meaning of all error codes through reversing.
right?

isage:
> ah got it. sorry, i thought wrongly. you mean, current error codes could be copied from ps4sdk or others. and it would be wrong. now, you check/dig meaning of all error codes through reversing. right?

yes

include/psp2/usbd.h
d3m3vilurr: can you explain a reason?

isage: enums are good for type-safety of func params, for example. but that isn't the case here. those are just "helpers" for human-readable values. defining them as enums would require either changing func and struct param types (which i'm against, since that would break struct compatibility between implementations on other systems) or would require type-casting each time they're used.
same goes for classes below.

d3m3vilurr:
merged 92776f3

PR #757, opened 2023-03-17
Fixed spelling error in struct SceFontNewLibParams in pgf.h

palaster:
Sce was spelled twice in the struct name instead of once. I am assuming since it was typedef with its correct name afterward this probably didn't make any real issue


d3m3vilurr:
thanks

PR #765, opened 2023-05-11
Added cpu atomic functions

Princess-of-Sleeping:


d3m3vilurr:
it needs to include `psp2common/types.h`

Princess-of-Sleeping:
done

PR #766, opened 2023-05-13
add pss_get_account_hash_id

KuromeSan:
under the android version of psm "libdefault.so" this is titled scePsmEdataGetAccountHashID however in that version it is not part of libmonobridge but rather directly in LibPsm; but the two functions look the same and are used for the same;

(they are the actual implementation of : [Sce.PlayStation.Services.AccountInformation.GetUniqueId](https://psmreborn.com/psm-sdk-doc/en/classSce_1_1PlayStation_1_1Core_1_1Services_1_1AccountInformation.html])


d3m3vilurr:
@KuromeSan please check the discord DM.

PR #777, opened 2023-06-17
Add MultiDisc signature generation functions, verify function, legacy psp generation and verify, and there vshbridge functions too

KuromeSan:
personally i find the ksceNpDrmPspEbootVerify and ksceNpDrmEbootSigVerify names very confusing, 

maybe should be like ksceNpDrmEbootSigVerifyLegacy and ksceNpDrmEbootSigVerify  instead- would be more descriptive

same with ksceNpDrmPspEbootSigGen and ksceNpDrmEbootSigGenPsp ..

just have ksceNpDrmEbootSigGenPspLegacy and ksceNpDrmEbootSigGenPsp. or something like that, the two very simular names is very confusing tbh

but anyway, heres the headers for them

### DIFF
--- a/include/psp2/vshbridge.h
+++ b/include/psp2/vshbridge.h
@@ -158,7 +158,7 @@ int _vshNpDrmEbootSigGenPs1(const char *eboot_pbp_path, const void *eboot_sha256
  *
  * @param[in]  eboot_pbp_path         - The pointer of the file path of the EBOOT.PBP file
  * @param[in]  sce_discinfo           - The pointer of contents of vs0:/app/NPXS10028/__sce_discinfo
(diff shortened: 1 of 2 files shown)


KuromeSan:
dont accept this yet, i got the vshbridge version of verifySig wrong, as it seems to take 3 arguments not 2..

KuromeSan:
so it like copies 8 bytes from arg3 from kernel space into userspace, and then does absoultely nothing with it after that,

hooking the function it seems to be a pointer to the same eboot_signature parameter again,

long** reserved; i guess idk

KuromeSan:
ok fixed now its all correct

PR #779, opened 2023-06-20
add SceNpDrmActivationData struct, and ksceNpDrmReadActData, and sceNpDrmCheckActData

KuromeSan:
+ the "SceNpDrmActivationData" struct is based on a simular struct from within NoPspEmuDrm: https://github.com/KuromeSan/NoPspEmuDrm/blob/master/user/PspNpDrm.h#L115, which is licensed under the GPLv3, but given that _i wrote this_ you have my permission to include this and license under MIT 

+ the PspAct struct in NoPspEmuDrm is itself based on the Act struct within Chovy-Sign2 just re written as C code https://github.com/KuromeSan/chovy-sign/blob/master/PspCrypto/SceNpDrm.cs#L317 that one was done by SquallATF, who also have no problem with it being included in VitaSDK. 

if its still somehow a problem i guess just change the output type of ksceNpDrmReadActivationData to void* 🤷‍♀️

### DIFF
--- a/include/psp2/npdrm.h
+++ b/include/psp2/npdrm.h
@@ -38,6 +38,21 @@ int _sceNpDrmGetRifName(char *rif_name, uint64_t aid);
 int _sceNpDrmGetFixedRifName(char *rif_name, uint64_t aid);
 
 /**
+ * Check you have npdrm activation data, and get information from it
+ *
+ * @param[out]  act_type        - The pointer of activation type output.
+ *
+ * @param[out]  version_flag    - The pointer of version flag output.
+ *
+ * @param[out]  account_id      - The pointer of activated account id output.
+ *
+ * @param[out]  act_exp_time    - The pointer of activation expire time output, [0] is start_date, [1] is end_date
+ *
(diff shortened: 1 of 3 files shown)


KuromeSan:
ok it should be good now xD

KuromeSan:
alright i think thats all of it..

d3m3vilurr:
@KuromeSan can you squash commits? or i'll just merge with squash every commits if you are okay.

KuromeSan:
How exactly to squash commit? :? idk if u can do it yourself then just do that i guess

KuromeSan:
Ah i worked it out, alright. should be good now

d3m3vilurr:
> How exactly to squash commit? :? idk if u can do it yourself then just do that i guess

imo, you already did it. but i explain more detail steps if you want to make multiple commits than just one :)
```bash
git stash -i HEAD~10 # use commit count
# it will open the $EDITOR then you can pick, squash, re-edit and re-ordering of the commits
# after this action you get one or more commits
git push -f origin target-branch # in this case `git push -f origin patch-19`
```

PR #786, opened 2023-07-03
Add deprecation messages guard

isage:


isage:
See https://github.com/vitasdk/vita-toolchain/pull/240

PR #795, opened 2023-07-18
Update nid dbs

Princess-of-Sleeping:
~~Do not merge this PR yet.~~ PR is ready.


vita.header_warn.cmake
```diff
@@ -48,6 +48,13 @@ endif()
 math(EXPR __vitasdk_expires_wrong_libs_module_name "20240101 + 2 * 100")
 
 if(${__vitasdk_current_time} LESS ${__vitasdk_expires_wrong_libs_module_name})
+endif()
+
+if(Off)
```
d3m3vilurr, line 53: hm... how devs read this message if you move this block to off part?

Princess-of-Sleeping: Disabled the message by correcting the module name without changing the lib name.

PR #815, opened 2023-09-24
last_built_toolchain.py: fetch 100 items instead of 30

isage:
currently last linux tag is beyond default 30 items limit, so raise it to max (100)
~~"proper" fix would be to iterate through pages until we get required tag (but that could trigger api rate limit)~~
added iteration through 5 first pages

### DIFF
--- a/.travis.d/last_built_toolchain.py
+++ b/.travis.d/last_built_toolchain.py
@@ -10,7 +10,7 @@ except ImportError:
 
 GITHUB_REPO = 'vitasdk/autobuilds'
 GITHUB_API = 'https://api.github.com'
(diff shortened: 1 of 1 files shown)


scribam (approved):

d3m3vilurr:
looks good to me. I just worried this patch trigging the api rate limit but this is good  on the normal situation.

imo, we should remove the old built tags when succeded the cronjob CI.

isage:
> imo, we should remove the old built tags when succeded the cronjob CI.

Maybe, but i'd keep X last tags anyway, makes testing for regressions easier

PR #820, opened 2023-10-12
Fixed lot of cpu stuff

Princess-of-Sleeping:
From https://github.com/vitasdk/vita-headers/blob/master/vita.header_warn.cmake#L43

This PR will be merged 2023/10/15 or after.

### DIFF
--- a/include/psp2kern/kernel/cpu/atomic.h
+++ b/include/psp2kern/kernel/cpu/atomic.h
@@ -1,0 +1,125 @@
+/**
+ * \kernelgroup{SceCpu}
+ * \usage{psp2kern/kernel/cpu/atomic.h}
+ */
+
+
+#ifndef _PSP2KERN_KERNEL_CPU_ATOMIC_H_
+#define _PSP2KERN_KERNEL_CPU_ATOMIC_H_
+
+#include <psp2kern/types.h>
+
+#ifdef __cplusplus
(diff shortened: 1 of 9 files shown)


include/psp2kern/kernel/cpu/atomic.h
```diff
@@ -0,0 +1,125 @@
...
+#define _PSP2KERN_KERNEL_CPU_ATOMIC_H_
+
+#include <psp2kern/types.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+int ksceKernelAtomicAdd32AndGet64InHiLoRange(unsigned long long *val, int add_val, int limit);
```
d3m3vilurr, line 17: just a question, don't we need to change types to `SceInt32` and `SceUInt64`??

Princess-of-Sleeping: The SceCpuForKernel::atomic stuff was added straight from the wiki. I can change it to SceInt32/SceInt64 if you wish.

d3m3vilurr: yep. can you change them?
i'll merge this after that jobs

Princess-of-Sleeping: done

PR #839, opened 2024-12-10
Add missing definitions for adhoc matching headers

german77:


include/psp2/net/adhoc_matching.h
```diff
@@ -0,0 +1,101 @@
...
+    SCE_NET_ADHOC_MATCHING_HANDLER_EVENT_DENY         = 4,
+    SCE_NET_ADHOC_MATCHING_HANDLER_EVENT_CANCEL       = 5,
+    SCE_NET_ADHOC_MATCHING_HANDLER_EVENT_ACCEPT       = 6,
+    SCE_NET_ADHOC_MATCHING_HANDLER_EVENT_ESTABLISHED  = 7,
+    SCE_NET_ADHOC_MATCHING_HANDLER_EVENT_TIMEOUT      = 8,
+    SCE_NET_ADHOC_MATCHING_HANDLER_EVENT_ERROR        = 9,
+    SCE_NET_ADHOC_MATCHING_HANDLER_EVENT_BYE          = 10,
+    SCE_NET_ADHOC_MATCHING_HANDLER_EVENT_DATA         = 11,
+    SCE_NET_ADHOC_MATCHING_HANDLER_EVENT_DATA_ACK     = 12,
+    SCE_NET_ADHOC_MATCHING_HANDLER_EVENT_DATA_TIMEOUT = 13
```
d3m3vilurr, line 64: - have you get these enums via reversing module by yourself?
- doesn't referenced leaked/official sdk?

german77: I reversed and re-implemented the full module in https://github.com/Vita3K/Vita3K/pull/3470 although my work is a continuation of another WIP from another user that happens to have this exact enum already. The module behavior does seem to match with the enum names.

Since I can't confirm the source of this enum I can remove it as it's not essential for this header to work properly.

d3m3vilurr: yep. please remove them and keep only your own reversing part. we cannot allow official sdk part. unknown source has same risk.

german77: done

d3m3vilurr: @german77 sorry. something mores.

1. are remain enums safe?
2. please match code convention
    - use hardtab for indentation
    - if you doesn't have special reason, please use `snake_case` for structure field & function arguments (code coventions)
        - for example; `optLen` to be `opt_len`

@Princess-of-Sleeping can you support this review if I late :'(

german77: Remaining enums are safe. I did the rest of this header by myself and I also have an usage example that I will upstream once this is merged. I will address the formatting issues in the following minutes.

german77: Just as a side note the files in the same folder as this one don't match the convention.

d3m3vilurr:
ah you should define doc group in [here](https://github.com/vitasdk/vita-headers/blob/master/docs/definitions.dox#L332-L336)

d3m3vilurr:
ty for your contribution & patient :)

PR #842, opened 2025-01-03
Added `const` keyword for vfs.h

Princess-of-Sleeping:
@bythos14 If you forgot this on prev PR, You agree for this?

Im sure on some member have `const`.


bythos14:
Oh yeah, I overlooked those when I exported the structs from Ghidra. Thanks for noticing

Princess-of-Sleeping:
OK, I think PR merge is ready now.

PR #844, opened 2025-01-30
add SceSdifCmd56 nids

LiEnby:
used to  communicate with MMC device over GENERAL_CMD, (used on vita for gamecart authentication)


Princess-of-Sleeping:
Fix a sort order.

PR #850, opened 2025-03-16
Fix docs error

bythos14:
Fix error preventing docs from generating with Modules section


Princess-of-Sleeping:
Superseded by PR #851

PR #852, opened 2025-04-05
db: remove duplicate entries

CreepNT:
There are duplicate entries in the following files:
- db/360/SceLibMonoBridge.yml
	* pss_errno_loc
	* pss_app_exit_liveboard
	* (both removed as unknown which one is correct)

- db/360/SceLibMtp.yml
	* sceMtpBeginHttpGetPropertyWithUrl
	* sceMtpBeginGetSystemSetting
	* (wrong name determined thanks to NID calculation)

- db/360/ScePaf.yml
	* scePafResourceGetAttributeIdInt
	* scePafResourceGetAttributeIdIntLpb
	* (changed NID of scePafResourceGetAttributeIdIntLpb to match NID of function calling scePafResourceGetIdIntLpb)

cc @LiEnby to fix file `#1`
cc @Princess-of-Sleeping to check `#3`

### DIFF
--- a/db/360/SceLibMonoBridge.yml
+++ b/db/360/SceLibMonoBridge.yml
@@ -184,7 +184,6 @@ modules:
           monoeg_try_realloc: 0xB1006CCC
           pss_alloc_mem: 0x1A2E441F
           pss_alloc_raw: 0x599AEA5E
-          pss_app_exit_liveboard: 0x58ABAD62
           pss_code_mem_alloc: 0xF466ABEC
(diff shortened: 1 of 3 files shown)


LiEnby:
Hm I believe android has symbols to LibMonoBridge, perhaps i can find the original name that way, it’s a thing I’ve been doing in SnowPME for awhile ..

Princess-of-Sleeping:
Confirmed.

PR #862, opened 2025-07-29
Added SceGxmTexture*AnisoMode

FirebirdTA01:


isage:
wdym "verify my NID calculation"
where did you even get those functions and NIDs from?
there's no  such nids nor functions in libgxm.

Rinnegatamante:
Assuming that NID existed (since it was referenced before even in vitasdk), it's long gone and probably present only in some very early fw (maybe even proto fw only).

isage:
There's no such NIDs in libgxm_es2, nor libgxm_es3, nor libgxm_es4 (from 0.99)

isage:
Functions themselves did exist, but only before 1.00 and with different NIDs (0xe719cbd4 for set, and de9d5911 for get)

FirebirdTA01:
> wdym "verify my NID calculation" where did you even get those functions and NIDs from? there's no such nids nor functions in libgxm.
> 
> you can't just take random names, calculate NID for them and expect them to exist.

I got them from old copies of my vita apps which used to use this function. Are there modern replacements or are all textures just automatically filtered when at extreme angles now?

bythos14:
@Rinnegatamante We did test this with GTA SA I believe, and there was no noticeable change as to indicate that the functionality was still present in the ES4 (retail) GPU revision, ergo why Sony removed it from gxm's API.

PR #865, opened 2025-09-24
Add SceError library

LiEnby:
found documented in vita3k emulator, as well as a plugin from princess-of-sleeping,

https://github.com/Vita3K/Vita3K/blob/556c2355a42517ad92bc9365ce6542b2aae39227/vita3k/modules/SceError/SceError.h#L20

https://github.com/Princess-of-Sleeping/vita-utility/blob/a9e689bb5ca0d74654833c305330dd2d6983b687/kernel/SceErrorHelper/src/main.c#L19

### DIFF
--- a/docs/definitions.dox
+++ b/docs/definitions.dox
@@ -301,6 +301,9 @@
  *     \defgroup SceCoredump coredump module
  *      Module for process crash handler
  *
+ *     \defgroup SceError error code module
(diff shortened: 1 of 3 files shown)


d3m3vilurr:
you can't reference from vita3k code(GPL). but you can ref princess code(MIT)

thanks :)

PR #886, opened 2026-07-30
Add and fix functions/structures based on wiki + re’ing

devnoname120:
This PR adds missing exported functions/structs/NIDS, etc.. This is based on a mix of Henkaku’s wiki and a lot of reverse engineering on dumped FW 3.60. I tried to keep backward compatibility for some of the stuff but it’s not 100% perfect. For the completely wrong struct member names (like they don’t even make sense because it has a completely different purpose/meaning) I broke compatibility but I think it’s for the greater good (and they are niche things so it shouldn’t break too many homebrews, if any)


db/360/SceSblPostSsMgr.yml
```diff
@@ -58,11 +63,22 @@ modules:
...
           ksceSblRtcMgrGetCpRtcLogical: 0xDE5150FE
           ksceSblRtcMgrGetCpRtcPhysical: 0x942010A0
+          ksceSblRtcMgrSetCpRtcLogical: 0xDC68EE5C
+          ksceSblSsMgrCloudDataGetEncDecCryptHandle: 0xC6684F7E
+          ksceSblSsMgrCloudDataGetSignCryptHandle: 0x739C981E
+          ksceSblSsMgrCloudDataStop: 0x2BF04B8E
           ksceSblUtMgrExecuteUtokenSmCommand1: 0xC2E58CE3
           ksceSblUtMgrGetTrilithiumBuffer: 0xABDD68CD
           ksceSblUtMgrHasComTestFlag: 0x7ACCAA50
+          ksceSblUtMgrHasFlag1: 0x22599675
```
devnoname120, line 72: Self-review: what are flags 1-9 and should we name them instead?

include/psp2/appmgr.h
```diff
@@ -566,22 +575,457 @@ VITASDK_BUILD_ASSERT_EQ(0x428, SceAppMgrCoredumpState);
...
+	SceUInt8 infoBarTransparency; //!< One of ::SceAppMgrInfoBarTransparency.
+	SceUInt8 reserved4B; //!< High byte of the packed info-bar state.
+	SceUInt32 reserved4C; //!< Copied from the process context; its purpose is unknown.
+	char bgmProxyAppName[32];
+	SceUInt8 processLaunchType; //!< FW 3.60 writes 0, 1, or 2; value 2 identifies a direct path launch.
+	                            //!< The purposes of 0 and 1 are unknown.
+	SceUInt8 applicationType; //!< One of ::SceAppMgrApplicationType.
+	SceUInt8 reserved72[2]; //!< Set to 0 on FW 3.60.
+	SceUInt32 appFlags;
+	SceUInt32 recommendedScreenOrientation; //!< A value from 1 through 4.
```
devnoname120: Self-review: create an enum for the screen orientations

scribam:
@devnoname120 can you rebase to master please? That should fix the "Install packages" step in CI and run the other steps

devnoname120:
@frangarcj Can you review?

frangarcj:
Really nice PR. One compatibility concern after looking through the larger struct/header changes: there are several places where an already-public struct member is renamed even though its size/offset stays the same. That preserves ABI, but it still breaks existing source code that accesses the old member name.

Would it make sense to preserve the old names as aliases (anonymous unions/structs), similar to what this PR already does in `modulemgr.h` for fields such as `unk_10` / `alignment` and `number_of_imported` / `client_count`?

For example, `SceSelfAuthInfo` could keep both names without changing layout:

```c
union {
    SceUInt64 program_sceversion;
    uint8_t padding[8]; /* legacy name */
};

union {
    SceSharedSecret shared_secret;
    SceSharedSecret secret; /* legacy name */
};
```

The same idea could be applied to pure renames in structs such as `SceAppMgrLaunchParam`, `SceKernelAllocMemBlockKernelOpt`, `SceClass`, `SceSblRsaPrivateKeyParam`, the FIOS syscall argument structs, etc. The new semantic names are definitely better; keeping the old names as deprecated/legacy aliases would just make the transition much less disruptive for existing homebrew.

It might also be worth adding a tiny compile-only source-compat regression test that references a few of the legacy member names. The current size/ABI/NID/sample checks are great, but they won't catch a member rename when the binary layout itself is unchanged.

There are of course genuine layout corrections in this PR (for example structs whose actual size changes), where aliases can't solve compatibility. My suggestion is mainly for the cases where the change is only a better member name and preserving the old spelling is effectively free.

devnoname120:
@frangarcj Thank you for the review! Problem with the union technique is that it can make things pretty unreadable instead of having flat structures you end up with a bunch of unions inside it.

I tried in general to keep some backward compatibility except where previous names didn’t make any sense/were completely incorrect and when the current APIs were never or almost never used (or when the usages were in homebrews that just embedded their own definition of structs because the ones from the vitasdk were incorrect). I’m not 100% sure I managed to do it, but I’m willing to submit PRs to all the vita homebrews/plugins using the affected APIs in order to make them compatible with the non backward-compatible changes.

What do you think?

devnoname120:
@scribam Done

include/psp2/io/fcntl.h
```diff
@@ -199,9 +201,118 @@ int sceIoSetPriority(SceUID fd, int priority);
...
+ * User export used by ::sceIoDevctl.
+ *
+ * @param[in] devname - Device name.
+ * @param[in] cmd - Device-specific command.
+ * @param[in] arg - Input buffer containing \a opt->arglen bytes, or NULL.
+ * @param[in] opt - Required block describing the input and output buffers.
+ *
+ * @return The device-driver result, or a negative error code.
+ */
+int _sceIoDevctl(const char *devname, int cmd, const void *arg, const SceIoDevctlOpt *opt);
```
devnoname120: Should I remove all the `_`? I think it would require editing the NIDs databases (afaik I just re-used the names as-is when they were already defined in the NIDs database

include/psp2/kernel/backtrace.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include <psp2common/kernel/backtrace.h>
+#include <psp2/types.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct SceBacktraceArgs {
+	SceUInt32 *pNumReturn; //!< Receives the number of frames written to the output buffer. May be NULL.
```
devnoname120: Self-review: change to `num_frames`, `frames_count` or similar

include/psp2/io/fcntl.h
```diff
@@ -199,6 +200,73 @@ int sceIoSetPriority(SceUID fd, int priority);
...
+SceUID _sceIoDopen(const char *dirname, sceIoDopenOpt *opt);
+int _sceIoDread(SceUID fd, SceIoDirent *dir);
+int _sceIoGetstat(const char *name, SceIoStat *buf, sceIoGetstatOpt *opt);
+int _sceIoGetstatByFd(SceUID fd, SceIoStat *stat);
+int _sceIoMkdir(const char *dirname, SceIoMode mode, sceIoMkdirOpt *opt);
+SceUID _sceIoOpen(const char *filename, int flag, SceIoMode mode, sceIoOpenOpt *opt);
+int _sceIoRemove(const char *filename, sceIoRemoveOpt *opt);
+int _sceIoRename(const char *oldname, const char *newname, sceIoRenameOpt *opt);
+int _sceIoRmdir(const char *dirname, sceIoRmdirOpt *opt);
+int _sceIoSync(const char *device, unsigned int flags, sceIoSyncOpt *opt);
```
d3m3vilurr: I'm not sure that we need these internal functions.
does this have benefit than `SceLibKernel` version??

or how about replace it as them? if these always needs the opt, we might give another public name (for example just `2`)

include/psp2/kernel/backtrace.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include <psp2common/kernel/backtrace.h>
+#include <psp2/types.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct SceBacktraceArgs {
+	SceUInt32 *pNumReturn; //!< Optional pointer that receives the number of frames gathered.
```
d3m3vilurr: snake_case?

include/psp2/kernel/backtrace.h
```diff
@@ -0,0 +1,45 @@
...
+ * buffer must have room for at least one frame.
+ * @param[in] pArgs Required backtrace parameters.
+ *
+ * @return With ::SCE_KERNEL_BACKTRACE_MODE_DONT_EXCEED, 0 on success.
+ * Otherwise, the call-stack depth on success. Returns < 0 on error.
+ *
+ * @note On FW 3.60 this function requires development mode. Kernel-mode
+ * unwinding also requires the corresponding QAF permission.
+ */
+SceInt32 _sceKernelBacktrace(SceUID threadId, SceKernelCallFrame *pCallFrameBuffer, SceSize numBytesBuffer, SceBacktraceArgs *pArgs);
```
d3m3vilurr: can you use snake_case for variable name?

I know, we already use mixed things, but keep snake_case if a file is first added.
other newer files are same.

include/psp2/kernel/processmgr.h
```diff
@@ -90,9 +90,29 @@ SceKernelTime sceKernelLibcTime(SceKernelTime *tloc);
...
+	int (*sceKernelExitThread)(int exitStatus);
+	int (*sceKernelExitDeleteThread)(int exitStatus);
+	int (*_sceKernelExitCallback)(void);
+	void *coredumpHandler; //!< A pointer to a function with signature `int (SceSize args, void *argp)`.
+	int *pProcessTime; //!< Points to the ::SceKernelSysClock process-time base.
+	int *pPMUSERENR; //!< Points to the cached ::SceUInt32 PMUSERENR value.
+} SceLibkernelAddresses;
+VITASDK_BUILD_ASSERT_EQ(0x1C, SceLibkernelAddresses); // size is from FW 3.60
+
+int _sceKernelExitProcessForUser(int status);
```
d3m3vilurr: it should be `sceKernelExitProcess` (remove `_` prefix and cut the `ForUser`)
but it's already existed in this file.

so if same purpose function is alive in headers, we don't need it.
please don't add mechanically ;)

include/psp2/kernel/processmgr.h
```diff
@@ -90,9 +90,29 @@ SceKernelTime sceKernelLibcTime(SceKernelTime *tloc);
 
 int sceKernelLibcGettimeofday(SceKernelTimeval *tv, SceKernelTimezone *tz);
 
+typedef struct SceLibkernelAddresses {
+	SceSize size; //!< Size of this structure
+	int (*sceKernelExitThread)(int exitStatus);
+	int (*sceKernelExitDeleteThread)(int exitStatus);
+	int (*_sceKernelExitCallback)(void);
```
d3m3vilurr: `kernel_exit_callback`, `kernelExitCallback`, `callbackForKernelExit`, ...

imho, these are just function pointer, we don't need `sce` prefix for these

include/psp2/kernel/sysmem.h
```diff
@@ -186,6 +171,44 @@ int sceKernelGetFreeMemorySize(SceKernelFreeMemorySizeInfo *info);
...
 
+typedef struct SceKernelSubbudgetInfo {
+	int size; //!< Size of this structure. Must be set to 0xC.
+	SceUInt32 totalSize; //!< Total size of the subbudget in bytes.
+	SceUInt32 freeSize; //!< Free size of the subbudget in bytes.
+} SceKernelSubbudgetInfo;
+VITASDK_BUILD_ASSERT_EQ(0xC, SceKernelSubbudgetInfo); // size is from FW 3.60
+
+#define SCE_KERNEL_SUBBUDGET_ID_MAIN (0)
+#define SCE_KERNEL_SUBBUDGET_ID_CDLG (1)
```
d3m3vilurr: how about define enum?

include/psp2/appmgr.h
```diff
@@ -33,6 +38,10 @@ typedef enum SceAppMgrSystemEventType {
 	SCE_APPMGR_SYSTEMEVENT_ON_STORE_REDEMPTION   = 0x10000006
 } SceAppMgrSystemEventType;
 
+#define SCE_APP_EVENT_ON_ACTIVATE   (0x10000001)
+#define SCE_APP_EVENT_ON_DEACTIVATE (0x10000002)
+#define SCE_APP_EVENT_REQUEST_QUIT  (0x20000001)
```
d3m3vilurr: it also needs enum. but... nobody use it on this commit?

include/psp2/dmac5.h
```diff
@@ -8,35 +8,30 @@
...
-	SceSize length;  //<! The src data length
-	const void *key; //<! The key data
-	SceSize keysize; //<! The key size in bits
-	void *ctx;       //<! SceSblDmac5HashTransformContext Or another context of size 0x10-bytes
+	const void *src; //!< The operation input buffer
+	void *dst;       //!< The operation output buffer
+	SceSize length;  //!< The src data length
+	const void *key; //!< The key data
+	SceSize keysize; //!< The key size in bits
+	void *ctx;       //!< A pointer to a ::SceSblDmac5HashTransformContext structure.
```
d3m3vilurr: these section can split commit and we can merge it on another PR.. but it's less important anyway.

include/psp2/dmac5.h
```diff
@@ -54,13 +49,13 @@ int sceSblDmac5EncDec(SceSblDmac5EncDecParam *param, SceUInt32 command);
...
- * @param[inout] param   - The encdec param.
+ * @param[inout] param   - The hash transform param.
  * @param[in]    command - The DMAC5 hash base command.
- * @param[in]    extra   - The DMAC5 extra command.
+ * @param[in]    flags   - Bitwise OR of ::SceSblDmac5HashFlag values.
  *
  * @return 0 on success, else < 0.
  */
-int sceSblDmac5HashTransform(SceSblDmac5HashTransformParam *param, SceUInt32 command, SceUInt32 extra);
+int sceSblDmac5HashTransform(SceSblDmac5HashTransformParam *param, SceUInt32 command, SceUInt32 flags);
```
d3m3vilurr: `..., SceSblDmac5HashFlag flags);`

include/psp2/error.h
```diff
@@ -7,14 +7,123 @@
...
+ * @param[out] error_string - Optional external error-code buffer. If non-NULL, it must point to at least 16 bytes.
  * @param[in] error_code - Internal error code
  *
  * @return 0 on success, < 0 on error.
  */
-int _sceErrorGetExternalString(char* error_string, int error_code);
- 
-#endif //_PSP2_ERROR_H_
\ No newline at end of file
+int _sceErrorGetExternalString(char *error_string, int error_code);
```
d3m3vilurr, line 34: `...(SceErrorStrings *error_string, ...);`

include/psp2/error.h
```diff
@@ -7,14 +7,123 @@
...
 #define _PSP2_ERROR_H_
 
+#include <vitasdk/build_utils.h>
+#include <psp2/types.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct SceErrorStrings {
```
d3m3vilurr: just `SceErrorString`?

include/psp2/touch.h
```diff
@@ -147,6 +148,17 @@ int sceTouchEnableTouchForce(SceUInt32 port);
...
+/**
+ * Get touch-panel device information.
+ *
+ * @param[in] panel One of ::SceTouchPortType.
+ * @param[out] pInfo The buffer to receive the device information.
+ *
+ * @return 0 on success, < 0 on error.
+ * @note This function is restricted to system programs.
+ */
+int sceTouchGetDeviceInfo(SceUInt32 panel, SceTouchDeviceInfo *pInfo);
```
d3m3vilurr: is panel different as port?

d3m3vilurr (commented):
ok... i give up a review when i read half of first commit :'(

i'll send a pm on discord about this review...

include/psp2/dmac5.h
devnoname120: This is what I did initially but it breaks C++ compilation. This argument is a bitmask and in C++ OR-ing promotes enums to integers, which causes a type error when passing for example `SCE_SBL_DMAC5_HASH_FLAG_INPUT_CONTEXT | SCE_SBL_DMAC5_HASH_FLAG_OUTPUT_CONTEXT` as an argument. C is less strict and accepts it but doesn’t provide type safety anyway.

Requiring the user to do a `(SceSblDmac5HashFlag)` type cast on the argument every time they OR flags is bad UX IMO and doesn’t provide any safety which makes it pointless.

include/psp2/kernel/sysmem.h
devnoname120: I agree and this is already done in a follow-up commit that is part of the PR: https://github.com/devnoname120/vita-headers/blob/1e35ef7218b3c709d7e5a7d4481008491e55aa31/include/psp2/kernel/sysmem.h#L174-L184

I would recommend reviewing the PR using the “Changed files” tab instead of commit by commit because there are lots of changes and improvements I made so reviewing commit by commit is redundant. This PR is meant to be squashed for merging but I don’t like to force-push branches hence why there are multiple commits.

https://github.com/vitasdk/vita-headers/pull/886/changes?w=1

include/psp2/dmac5.h
devnoname120: I agree in principle but a later commit of the PR renames some fields and changes these comments, which means that it’s not solely a doxygen syntax fix but a semantic change that is part of the broader changes of this PR.

include/psp2/error.h
devnoname120: Problem is that it would cause a backward compatibility issue because this is not a new API. So someone who passes a char array would have the following compilation error in his project:

```
no known conversion from 'char[16]' to 'SceErrorStrings *'
```

include/psp2/kernel/backtrace.h
devnoname120: I will fix it, thanks for the comment

include/psp2/kernel/backtrace.h
devnoname120: I will fix it, thanks for the comment

include/psp2/error.h
devnoname120: Good point, I will change it

include/psp2/appmgr.h
devnoname120: They were moved to an enum in a later commit of the PR and they are used in the `SceAppMgr` struct, see the doxygen comments here: https://github.com/vitasdk/vita-headers/blob/1e35ef7218b3c709d7e5a7d4481008491e55aa31/include/psp2/appmgr.h#L716-L721

It’s defined as an `int` for compatibility reasons but it’s indeed of these types.

include/psp2/touch.h
devnoname120: No you’re right it’s the same. I agree that it should be either changed to `port_type` or the type should be changed to `SceTouchPanel` for consistency. I’m not sure where the port terminology comes from (IMO it’s confusing compared to `panel`) but the first option would improve consistency while avoiding to break compatibility so I will apply the change you requested.

include/psp2/kernel/processmgr.h
devnoname120: Is it the convention that we use in the vitasdk? I’m confused why struct types should have the `sce` prefix but function types should not have the `sce` prefix. IMO we should use the same convention for both in order to keep consistency.

include/psp2/kernel/processmgr.h
devnoname120: IMO it’s better to have known prototypes for functions than having NIDs and let the user wonder what the prototype of these functions is when he needs to hook them, define their types in IDA/Ghidra, etc.

Just the fact that they have the same prototype in `SceLibKernel` and `SceProcessmgr` is useful information IMO and keeping it out would reduce the information given to the developer, especially of plugins. Using different names in both headers makes sure that we don’t have a redeclaration problem if both are included. I could create a new header in `psp2common` just for the headers that are the same in `SceLibKernel` and `SceProcessmgr` but IMO it would add to confusion for developers to have some functions defined in these headers and other functions defined in this common ad-hoc header. This would add complexity to the doxygen documentation and make it less readable IMO.

include/psp2/io/fcntl.h
devnoname120: They are lower-level and we need them so that the developer knows their shape if hooking is necessary, import the headers in IDA/Ghidra to have all the reverse engineering information, etc. IMO it’s important that developers have all the information and arguments about functions in the NIDs yml databases, not just their names even if there is overlapping or duplicated functionality.

In this case there are also some behavioral differences between them but reverse engineering all differences and documenting them all would be a lot of additional reverse engineering so I’d rather not dig into each of them to write comments about the exact differences.

For example the `sceKernelExitProcess` of `SceLibKernel` compared to the one from `SceProcessmgr`:

* Uses an atomic guard to coordinate process exit
* Calls `sceKernelCallModuleExit(128)` before invoking `sceKernelExitProcess` from `SceProcessmgr`.
* Takes a different path through `sceKernelExitThread` when exit is already in progress

So these functions are not just duplicates of each other, they have different behavior. The one from `SceProcessmgr` is lower level whereas the one from `SceLibKernel` has additional convenience.

include/psp2/kernel/processmgr.h
devnoname120: The `_` prefix has two purposes for Sony as far as I know based on my reverse engineering and reading the wiki:

1) Mark some user functions as meant for internal use (not to be used by game developers). It’s just a convention because developers can still use them anyway (as long as they know the NID + prototype) but I suppose that as a rule of thumb Sony probably doesn’t document them or provide headers. I don’t know 100% because I don’t have access to the official SDK and I don’t want access due to copyright and intellectual property.<br>
  For these I think that it entirely makes sense to strip out the `_` prefix because we want to use the functions and whether Sony documented them or gave developers their blessing to use them or not doesn’t matter to us because we document them all anyway and many of them are useful for the users of the vitasdk.

2) Differentiate between low-level and high-level functions of similar purpose. Generally the `_` are lower-level and provide more options (for example providing a PID whereas the function without `_` automatically gets the current process PID and then calls the `_` version with that PID). This makes it cleaner and more convenient for the developer when low-level operations are not needed to be exposed to them (because e.g. they will only need to use these functions for current PID whereas `SceShell` may need to use it for other PIDs. This just is just an example, `_` difference is not always the PID).<br>
  Removing the `_` prefix and adding a `2` suffix removes this information and makes it unclear which is lower level and which is higher level. Sometimes the lower-level has more options, sometimes different behavior, etc. I prefer the `_` convention in this case because it gives the developer a clear understanding of which is which whereas `2` loses the hierarchical distinction entirely. Also Sony sometimes uses `2` for ABI change without breaking backward compatibility so again stripping `_` and putting `2` everywhere loses the distinction entirely and waters down the semantics.

include/psp2/kernel/backtrace.h
devnoname120: I just edited the whole PR to use snake_case for new things, unless they are part of existing structures in which case I followed the already-used convention for consistency. I will create an RFC to do a proposal for a broader rename of the vitasdk codebase and define precise rules. The rename will be done in a follow-up PR and for now I’d advocate to keep consistency.

include/psp2/kernel/backtrace.h
devnoname120: I just edited the whole PR to use snake_case for new things, unless they are part of existing structures in which case I followed the already-used convention for consistency. I will create an RFC to do a proposal for a broader rename of the vitasdk codebase and define precise rules. The rename will be done in a follow-up PR and for now I’d advocate to keep consistency.

include/psp2/kernel/processmgr.h
d3m3vilurr: because `sceFunction` are really sdk function, but function pointer of the structure is not a sdk function.
it can be user defined function.

that this reason I think I don't need `sce` prefix in here.

include/psp2/dmac5.h
d3m3vilurr: i mean. i just said bug fixing point that `//<!` change to `//!<`

(i'm not meaning `SceSblDmac5HashTransformContext ` part :))

include/psp2/error.h
d3m3vilurr: oh... first my review was wrong. it might be... `SceErrorString error_string` not a pointer.

but yeah... i agree that this can make breaking change.

include/psp2/kernel/processmgr.h
devnoname120: This is a very good point, I edited the PR accordingly

include/psp2/error.h
devnoname120: @d3m3vilurr if it’s not defined as a pointer then the `char[16]` will be passed by value, which means that the caller will never receive the error string.

An alternative would be to define it as a `typedef`:

```c
typedef char SceErrorString[16];

int _sceErrorGetExternalString(SceErrorString error_string, int error_code);
```

This would work because now it would pass by-value a pointer. However I don’t like it because it obscures the fact that `error_string` is the return value whereas `char *` makes it clear that it is a pointer rather than passing a value.

Another alternative would be to directly put `char[16]` as the argument type:

```c
int _sceErrorGetExternalString(char error_string[16], int error_code);
```

This would work because for array parameters both the C and C++ compilers erase the size from the type so they interpret it exactly as if it were just a raw pointer `char *error_string`:

```c
int _sceErrorGetExternalString(char error_string[16], int error_code);

char tiny[1];
_sceErrorGetExternalString(tiny, code); // valid: no compiler error despite the buffer overflow
```

So effectively no type safety for the function call: `[16]` gives devs the impression that the compiler enforces something that it doesn’t (it’s only cosmetic).

frangarcj:
I think the split makes sense, but I’d prefer this to be two separate PRs rather than two commits in the same PR: one PR containing only the backward-compatible additions/fixes, and a second PR containing the intentionally breaking API/ABI corrections on top of it.

That way we can review and merge the compatible part independently, and treat the breaking changes explicitly with their own discussion/testing/migration notes.

frangarcj:
Also, as soon as you open the backward-compatible part as its own PR, I’m happy to approve it.

